PR #192, opened 2024-03-12
generic Form which populates inputs based on the Model type bound to it

StoyanGoranov:
Although not functional at the moment the concept proves promising to prevent many similar configurations of Forms within the application


achobanov:
The name of the branch doesn't make much sense. I'd recommend you create branches for the smaller (task) issues. Also including a prefix for the target branch is a good idea. E.g. `v2/competition` or `v2/setup-competition`.

Issue #170, opened 2023-10-12
Create Setup UI


achobanov:
## Overview
Setup is the domain boundary responsible for the configuration of a single Event. An Event can span multiple days and usually has more than one Competition running simultaneously. 

## Domain Entities
- [x] Event
- [x] Official
- [x] #190 
- [x] #191
- [x] #206
- [x] #207
- [x] #222
- [x] #223
- [x] #210

Entities should implement `ISummarizable` . The idea is to call `Sumarize` on the root `Event` entity, which will render a tree-like graph for the entire Setup configuration, so that a user can check for any configuration errors. 

## Domain Objects
- Club - Clubs are the same for a given Country. The idea is to persist them across Events so that they don't have to be created every time. This mechanism isn't yet decided upon, so you can hardcode several clubs similarly to `Country`.

## NTS.Judge.Blazor
- [x] Page layout: #189 

Additionally add:
- `*Form.razor` - entity form used for both create and update. The trickiest part here is to get the validation to work, see `EventForm.RegisterValidationInjectors`
- `*FormModel.cs`  
- `*Page.razor` - entity page component for update

for every entity. Then add grid components for their direct children, same as `OfficialGrid.razor`

## NTS.Judge
Add `*Behind.cs` classes for every entity. Use `Event` and `Official` as example. 

##  Diagram
This is the Setup domain segment of the [v2 Domain model diagram](https://github.com/Not-Endurance/not-timing-system/blob/dev/v2/diagrams/NTS%20v2%20Domain%20Model.drawio.png?raw=true)

![image](https://github.com/Not-Endurance/not-timing-system/assets/1363532/7808e66f-76b5-4935-8884-43eab4ef7a51)

PR #195, opened 2024-03-25
Constestant and Loop - Setup Domain Entity

StoyanGoranov:
Related to: https://github.com/Not-Endurance/not-timing-system/issues/191


src-v2/Judge/NTS.Judge/Events/CompetitionChildrenBehind.cs
```diff
@@ -0,0 +1,45 @@
...
+using System.Threading.Tasks;
+
+namespace NTS.Judge.Events;
+public class CompetitionChildrenBehind : INotBehindParent<Contestant>
+{
+    private readonly IRepository<Competition> _competitionRepository;
+    private readonly IParentRepository<Contestant> _contestantRepository;
+    private Competition _competition;
+
+    public CompetitionChildrenBehind(IParentRepository<Contestant> contestants)
```
StoyanGoranov, line 17: The following feels like it must be conceptually wrong but I seem to lack the knowledge to reason why.
 ```C#
   public CompetitionChildrenBehind(IRepository<Competition> competition, IParentRepository<Contestant> contestants, Competition competitionParent)
   {
       _competitionRepository = competition;
       _contestantRepository = contestants;
   }
```

File: src-v2/NTS.Persistance/State.cs [hunk 2]
```diff
@@ -7,14 +8,26 @@ public class State :
...
+
+    public Competition? Competition
+    {
+        get => _competitions.FirstOrDefault();
+        set => _competitions = value == null ? new() : new() { value };
+    }
     public List<Official> Officials => Event?.Officials.ToList() ?? new();
     public List<Competition> Competitions => Event?.Competitions.ToList() ?? new();
 
+    public List<Contestant> Contestants => Competition?.Contestants.ToList() ?? new();
```
achobanov, line 29: Is this correct?

src-v2/NTS.Persistance/State.cs
```diff
@@ -7,14 +8,26 @@ public class State :
...
         set => _events = value == null ? new() : new() { value };
     }
+
+    public List<Tandem> Tandems { get; } = new();
+
+    public Competition? Competition
+    {
+        get => _competitions.FirstOrDefault();
+        set => _competitions = value == null ? new() : new() { value };
+    }
```
achobanov, line 25: Is this necessary?

src-v2/Not/Not.Blazor/TM/Navigation/NotNavigator.cs
```diff
@@ -33,7 +33,7 @@ public T ConsumeParameter<T>()
             throw GuardHelper.Exception($"Cannot get parameter '{typeof(T)}'. There are no parameters on this landing");
         }
         var result = _parameters.Get<T>();
-        _parameters = null;
+        //_parameters = null;
```
achobanov, line 36: Should this exist?

achobanov (changes requested):
There are warnings that should be addressed in Contestant.cs

There are bugs with Loop update and loop create operations. I suggest that you always check the full functionality before submitting a PR for review

achobanov:
Furthermore, there was no task and issue created for Loop implementation. I've created #206. You should be committing to the sub-task issues, rather than the big umbrella (#170) as it is too big to provide a useful context when looking at a given commit. Also the PR description should include `closes #` for both subtasks as we want Github to close them automatically as they hit master.

src-v2/Judge/NTS.Judge.Blazor/Pages/Setup/Competitions/CompetitionForm.razor
```diff
@@ -62,6 +62,6 @@
     public override Competition SubmitUpdate()
     {
         DateTimeOffset startTimeOffset = CombineStartDayAndTime(_model.StartDay, _model.StartTime);
-        return Competition.Update(_model.Name, _model.Type, startTimeOffset);
+        return Competition.Update(_model.Id, _model.Name, _model.Type, startTimeOffset);
     }
 }
```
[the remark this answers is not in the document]
StoyanGoranov: It isn't since StartDate isn't a property of the Domain Entity but it still seems like it needs to be validated somehow.

src-v2/Judge/NTS.Judge.Blazor/Pages/Setup/Contestants/ContestantGrid.razor
```diff
@@ -0,0 +1,42 @@
...
+@using Not.Services;
+
+@inherits NotComponent
+
+@if (_parentBehind.Children.Any())
+{
+    <NotH2 Class="pt-6 pl-6">@Localizer.Get("Contestants")</NotH2>
+    <NotButtonCreate Class="mt-1 ml-6 mb-3" OnClick="_contestantNavigator.Create" />
+    <MudDivider />
+    <MudDataGrid  T="Contestant" Items="_parentBehind.Children" ReadOnly="true" Bordered="true" Dense="true">
```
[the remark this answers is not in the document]
StoyanGoranov: ToString is indeed lighter and also saves time but my concern is it ranges from very user unfriendly to mildly user unfriendly depending on the implementation of the method as opposed to a table which is easier on the eyes.

src-v2/NTS.Domain.Setup/Entities/Loop.cs
```diff
@@ -1,19 +1,63 @@
...
+
+        Distance = distance;
+		Recovery = rec;
+		Rest = rest;
+		IsFinal = isFinal;
 	}
 
     public double Distance { get; private set; }
-	public int Rec { get; private set; }
+	public int Recovery { get; private set; }
```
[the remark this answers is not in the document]
StoyanGoranov: Added a semantically complete property name because it makes the code more readable - when one reads it, it gives information about its purpose and avoids a certain level of ambiguity.

src-v2/Judge/NTS.Judge.Blazor/Pages/Setup/Competitions/CompetitionForm.razor
```diff
@@ -35,7 +35,7 @@
     public DateTimeOffset CombineStartDayAndTime(DateTime? startDay, TimeSpan? startTime)
     {
         DateTime start = startDay.GetValueOrDefault(DateTime.MinValue);
-        DateTime startDayTime = start.Add((TimeSpan)startTime);
+        DateTime startDayTime = start.Date.Add((TimeSpan)startTime);
```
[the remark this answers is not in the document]
StoyanGoranov: GetValueOrDefault is a nice way to avoid an error when the nullable values are null but it will break the functionality of the application since the values are used within the logic. I think it is best to throw an error if the respective inputs contain null values.

src-v2/Judge/NTS.Judge.Blazor/Pages/Setup/Competitions/CompetitionFormModel.cs
```diff
@@ -16,13 +17,15 @@ public CompetitionFormModel()
 
     public CompetitionFormModel(Competition competition)
     {
+        Id = competition.Id;
         Name = competition.Name;
         Type = competition.Type;
         DateTime? startDay = competition.StartTime.DateTime;
         TimeSpan? startTime = competition.StartTime.DateTime.TimeOfDay;
         StartDay = startDay;
         StartTime = startTime;
```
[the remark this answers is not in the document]
StoyanGoranov: Yes, this is part of the logic that separates the Date and the Time of the Competition in separate inputs. If removed the functionality breaks down.

src-v2/Judge/NTS.Judge.Blazor/Pages/Setup/Competitions/CompetitionForm.razor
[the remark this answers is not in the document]
StoyanGoranov: Unsure, which project do you reference by the name 'Not.Common'. Am I missing something?

src-v2/Judge/NTS.Judge.Blazor/Pages/Setup/Competitions/CompetitionForm.razor
```diff
@@ -26,24 +27,17 @@
...
-        DateTimeOffset offsetTime = timeWithSpecifiedKind;
-        return offsetTime;
-    }
     public DateTimeOffset CombineStartDayAndTime(DateTime? startDay, TimeSpan? startTime)
     {
         DateTime start = startDay.GetValueOrDefault(DateTime.MinValue);
-        DateTime startDayTime = start.Date.Add((TimeSpan)startTime);
-        DateTimeOffset startTimeOffset = ToDateTimeOffSet(startDayTime);
+        DateTime startDayTime = start.Date.Add(startTime.GetValueOrDefault(TimeSpan.Zero));
+        DateTimeOffset startTimeOffset = startDayTime.ToDateTimeOffset();
```
achobanov, line 34: I think we should either keep them nullable or set default values on the model object.

src-v2/Judge/NTS.Judge.Blazor/Pages/Setup/Competitions/CompetitionUpdate.razor
```diff
@@ -14,7 +15,10 @@
     <NotFormUpdate T="Competition" TForm="CompetitionForm" Entity="_competition" UpdateAction="Update" />
 </MudPaper>
 
-<LoopGrid />
+<NotChildren Name="@Localizer.Get("Loops")" T="Loop" TForm="LoopForm"
```
achobanov, line 18: Localizer can be used internally in the component itself.

src-v2/Judge/NTS.Judge.Blazor/Pages/Setup/Contestants/ContestantForm.razor
```diff
@@ -3,29 +3,32 @@
...
 @inherits NotForm<Contestant>;
 
 <EditForm Model="_model">
-    <MudTimePicker DisableToolbar="true" @ref="_timeField" @bind-Time="_model.StartTime" Label="@Localizer.Get("Time")" />
+    <MudSwitch T="bool" @ref="_unrankedField" @bind-Value="_model.Unranked" Label="@Localizer.Get("Unranked")"
+               ThumbIcon="@(_model.Unranked==true ? Icons.Material.Filled.Done : Icons.Material.Filled.Close)"
+               ThumbIconColor="@(_model.Unranked==true ? Color.Success : Color.Error)" />
+    <MudSwitch T="bool" @bind-Value="_overrideStartTime" Label="@Localizer.Get("Override Start Time")"
+               ThumbIcon="@(_overrideStartTime==true ? Icons.Material.Filled.Done : Icons.Material.Filled.Close)"
+               ThumbIconColor="@(_overrideStartTime==true ? Color.Success : Color.Error)" />
```
achobanov, line 16: The switch configurations seem very similar. Can't we standardize them?

src-v2/Judge/NTS.Judge.Blazor/Pages/Setup/Contestants/ContestantForm.razor
```diff
@@ -35,13 +38,13 @@
 
     public override Contestant SubmitCreate()
     {
-        DateTimeOffset newStartTimeOffset = ToDateTimeOffSet(_model.StartTime);
-        return Contestant.Create(newStartTimeOffset);
+        DateTimeOffset? newStartTimeOffset = _model.StartTimeOverride.ToDateTimeOffset();
```
achobanov, line 41: Use `var newStartTimeOffset` instead of declaring the type explicitly,

src-v2/Judge/NTS.Judge.Blazor/Pages/Setup/Contestants/ContestantForm.razor
```diff
@@ -35,13 +38,13 @@
...
+        return Contestant.Create(newStartTimeOffset,_model.Unranked);
     }
 
     public override Contestant SubmitUpdate()
     {
-        DateTimeOffset newStartTimeOffset = ToDateTimeOffSet(_model.StartTime);
-        return Contestant.Update(_model.Id ,newStartTimeOffset);
+        DateTimeOffset? newStartTimeOffset = _model.StartTimeOverride.ToDateTimeOffset();
```
achobanov, line 47: Use implicit var declaration.

src-v2/Judge/NTS.Judge.Blazor/Pages/Setup/Contestants/ContestantForm.razor
```diff
@@ -35,13 +38,13 @@
...
+        return Contestant.Update(_model.Id ,newStartTimeOffset,_model.Unranked);
```
achobanov, line 48: Parameters should be separated by a comma and a space - `, `

src-v2/Judge/NTS.Judge.Blazor/Pages/Setup/Contestants/ContestantFormModel.cs
```diff
@@ -11,15 +11,24 @@ public class ContestantFormModel
...
+        StartTimeOverride = null;
     }
     public ContestantFormModel(Contestant contestant)
     {
         Id = contestant.Id;
-        TimeSpan? startTime = contestant.StartTimeOverride.DateTime.TimeOfDay;
-        StartTime = startTime;
+        if (contestant.StartTimeOverride != null)
+        {
+            TimeSpan? startTime = contestant.StartTimeOverride.Value.DateTime.TimeOfDay;
```
achobanov, line 21: This isn't correct. I'm not sure exactly what time this will return, but probably it will  simply be the time of day for the time-zone it was created in. So if it was in 6:00 in Sofia time then it could simply return 6:00 when you open it up in UK. You need to convert the DateTimeOffset value to the current timezone. This isn't a big issue right now, but eventually - when we have web APIs and databases it could become one.

achobanov: Also implicit vars. Explicit declaration is only acceptable if it is necessary.

src-v2/Judge/NTS.Judge.Blazor/Pages/Setup/Contestants/ContestantFormModel.cs
```diff
@@ -11,15 +11,24 @@ public class ContestantFormModel
...
+            StartTimeOverride = startTime;
```
achobanov, line 22: Also no need to create a variable and then assign it to the property. simply do
```
StartTimeOverride = contestant.StartTimeOverride....;
```

src-v2/Judge/NTS.Judge.Blazor/Pages/Setup/Contestants/ContestantFormModel.cs
```diff
@@ -11,15 +11,24 @@ public class ContestantFormModel
...
+        }
+        else
+        {
+            StartTimeOverride = null;
+        }
```
achobanov, line 27: Nullable definitions contain null value by default, it's how the runtime works, this expression is completely redundant.

src-v2/Judge/NTS.Judge.Blazor/Pages/Setup/Contestants/ContestantFormModel.cs
```diff
@@ -11,15 +11,24 @@ public class ContestantFormModel
...
+        Unranked = contestant.Unranked;
     }
 
     public int Id { get; set; }
-    public TimeSpan? StartTime { get; set; }
+    public TimeSpan? StartTimeOverride { get; set; }
+    public Boolean  Unranked { get; set; }
```
achobanov, line 33: Bool properties should indicate it in their name by posing a question. `IsUnranked` or `IsNotRanked` would be better.

src-v2/Judge/NTS.Judge.Blazor/Pages/Setup/Contestants/ContestantGrid.razor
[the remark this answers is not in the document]
achobanov: Do we still need this file? Why don't you replace it with `NotChildren`?

src-v2/Judge/NTS.Judge.Blazor/Pages/Setup/Events/EventPage.razor
```diff
@@ -35,7 +35,7 @@
         var entity = await _behind.Read(default);
         if (entity == null)
         {
-            throw GuardHelper.Exception($"Event cannot be null here");
+           GuardHelper.ThrowIfNull(entity, "Event cannot be null here");    
         }
```
achobanov, line 39: Does this make sense to you? Did you look inside the definition of `ThrowIfNull`?

src-v2/Judge/NTS.Judge.MAUI/wwwroot/css/app.css
```diff
@@ -3,6 +3,19 @@
...
+}
+
+.half-width{
+    width:45%;
+}
+
+.children-component{
+    margin-top:10px;
+}
+
```
achobanov, line 18: Can't we use Mud's styling framework instead of custom styles?

src-v2/Judge/NTS.Judge/Events/CompetitionChildrenBehind.cs
```diff
@@ -9,18 +9,17 @@ public class CompetitionChildrenBehind : INotBehindParent<Contestant>, INotBehin
     private readonly IRead<Competition> _competitionReader;
     private readonly IParentRepository<Contestant> _contestantParentRepository;
     private readonly IParentRepository<Loop> _loopParentRepository;
-    private Competition? _competition;
+    public Competition? _competition;
```
achobanov, line 12: Why public? Fields should always be private. Underscore also should only prefix private fields/properties.

src-v2/Judge/NTS.Judge/Events/ContestantChildrenBehind.cs
```diff
@@ -3,7 +3,7 @@
 
 namespace NTS.Judge.Events;
 
-public class ContestantChildrenBehind : INotBehindWithChildren<Contestant>
+public class ContestantChildrenBehind
```
achobanov, line 6: Do we need this class?

src-v2/NTS.Domain.Setup/Entities/Competition.cs
```diff
@@ -14,34 +14,20 @@ public class Competition : DomainEntity, ISummarizable, IParent<Contestant>, IPa
...
-        StartTime = startTime;
     }
 
     private Competition(string name, CompetitionType type, DateTimeOffset startTime)
     {
-        if (type == 0)
+        if (type == default)
         {
-            throw new DomainException(nameof(Type), "Competition type must have a value different from 0.");
+            throw new DomainException(nameof(Type), "Competition Type cannot be default");
```
achobanov, line 26: This is far better than "different from 0", but why not "Competition type is required"?

src-v2/NTS.Domain.Setup/Entities/Contestant.cs
```diff
@@ -6,37 +6,32 @@
...
-        }
         Id = id;
-        StartTimeOverride = startTimeOverride;
     }
-    private Contestant(DateTimeOffset startTimeOverride)
+    private Contestant(DateTimeOffset? startTimeOverride, bool unranked)
     {
-
-        if (startTimeOverride.DateTime.CompareTo(DateTime.Today) < 0)
+        if (startTimeOverride!=null && startTimeOverride.Value.DateTime.CompareTo(DateTime.Today) < 0)
```
achobanov, line 23: Space around operators.

src-v2/NTS.Domain.Setup/Entities/Loop.cs
```diff
@@ -5,55 +5,33 @@ namespace NTS.Domain.Setup.Entities;
 
 public class Loop : DomainEntity, ISummarizable
 {
-    public static Loop Create(double distance, int rec, int rest, bool isFinal) => new(distance, rec, rest, isFinal);
+    public static Loop Create(double distance, int rec, int rest) => new(distance, rec, rest);
```
achobanov, line 8: rec -> recovery

src-v2/NTS.Domain.Setup/Entities/Loop.cs
```diff
@@ -5,55 +5,33 @@ namespace NTS.Domain.Setup.Entities;
...
 
-    public static Loop Update(int id, double distance, int rec, int rest, bool isFinal) => new(id, distance, rec, rest, isFinal);
+    public static Loop Update(int id, double distance, int rec, int rest) => new(id, distance, rec, rest);
```
achobanov, line 10: rec -> recovery

src-v2/NTS.Domain.Setup/Entities/Official.cs
```diff
@@ -17,9 +17,9 @@ private Official(int id, Person person, OfficialRole role)
     }
     private Official(string name, OfficialRole role)
     {
-        if (role == 0)
+        if (role == default)
         {
-            throw new DomainException(nameof(Role), "Official Role must have a value different from 0.");
+            throw new DomainException(nameof(Role), "Official Role cannot be default");
```
achobanov, line 22: required

src-v2/Not/Not.Blazor/TM/NotChildren.razor
```diff
@@ -0,0 +1,62 @@
...
+<MudPaper Elevation="0" Class="children-component half-width">
+@if (_parentBehind.Children.Any())
+{
+    <NotH2 Class="pt-6 pl-6">@Name</NotH2>
+    <NotButtonCreate Class="mt-1 ml-6 mb-3" OnClick="_contestantNavigator.Create" />
+    <MudDivider />
+    @foreach (var item in Items)
+    {
+        <p class="mt-1 ml-6 mb-1">
+            @item.ToString()
```
achobanov, line 20: Think `ToString()` will be called implicitly by the blazor framework. Try it and remove it if isn't necessary.

src-v2/Not/Not.Blazor/TM/NotChildren.razor
```diff
@@ -0,0 +1,62 @@
...
+        </div>
+    </MudPaper>
+}
+</MudPaper>
+
+@code {
+    [Parameter]
+    public string Name { get; set; } = default!;
+    [Parameter]
+    public string NoItemsMessage { get; set; } = default!;
```
achobanov, line 45: This message is going to be the same for every entity. Why not simply construct it using the `Name` parameter and a string constant?

src-v2/Not/Not.Blazor/TM/NotChildren.razor
```diff
@@ -0,0 +1,62 @@
...
+    {
+        Items = _parentBehind.Children;
+    }
+
+    [Inject]
+    private INotBehindParent<T> _parentBehind { get; set; } = default!;
+    [Inject]
+    private IFormNavigator<T, TForm> _contestantNavigator { get; set; } = default!;
+    [Inject]
+    private INavigator _navigator { get; set; } = default!;
```
achobanov, line 61: Is this used?

src-v2/Not/Not/Extensions/DateTimeOffsetExtension.cs
```diff
@@ -0,0 +1,25 @@
...
+namespace Not.Extensions;
+public static class DateTimeOffsetExtension
+{
+    public static DateTimeOffset ToDateTimeOffset(this DateTime time)
+    {
+        DateTime timeWithSpecifiedKind = DateTime.SpecifyKind(time, DateTimeKind.Local);
+        DateTimeOffset offsetTime = timeWithSpecifiedKind;
+        return offsetTime;
+    }
+    public static DateTimeOffset? ToDateTimeOffset(this TimeSpan? timeToBeAdded)
```
achobanov, line 16: I don't think we should involve nullability in extensions, unless we're extending `Nullable` itself. Instead you should make sure the method is not executed on null value. See [null-conditional operator](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/member-access-operators#null-conditional-operators--and-)

achobanov (commented):
Overall pretty good progress in the right direction. Mostly tiding up to do :) Good job

achobanov (changes requested):
Still required tho :P

src-v2/Not/Not.Blazor/TM/NotChildren.razor
StoyanGoranov: Kept the parameter for custom messages but now setting the NoItemsMessage as a default.

src-v2/Judge/NTS.Judge.MAUI/wwwroot/css/app.css
StoyanGoranov: Some styles can be set with Mud styling.
Like `margin-top:10px;` but the general idea here is to set a class which can target an entire component or apply styles that are missing from Mud as a semantic property, e.g. `half-width` should ensure that two components with the class will fit on the same line.
I suggest we keep these class configurations until the focus comes to the UI fine tuning.

src-v2/Judge/NTS.Judge.Blazor/Pages/Setup/Competitions/CompetitionForm.razor
StoyanGoranov: Not sure I understand what you mean in this comment.

src-v2/Judge/NTS.Judge.Blazor/Pages/Setup/Competitions/CompetitionUpdate.razor
```diff
@@ -10,17 +10,20 @@
...
-<NotChildren Name="@Localizer.Get("Loops")" T="Loop" TForm="LoopForm"
-    UpdateRoute="@LOOP_UPDATE"
-    NoItemsMessage="@Localizer.Get("No loops have been created for this event.")">
-</NotChildren>
-
-<ContestantGrid />
+<MudPaper Class="d-flex justify-space-between pr-6" Elevation="0">
+    <NotChildren Name="Loops" T="Loop" TForm="LoopForm"
+        UpdateRoute="@LOOP_UPDATE">
+    </NotChildren>
```
achobanov, line 21: When defined like this the component looks as if UpdateRoute is it's content. Why is it on a new line and why not self-closing component?

src-v2/Judge/NTS.Judge.Blazor/Pages/Setup/Contestants/ContestantFormModel.cs
```diff
@@ -18,17 +18,12 @@ public ContestantFormModel(Contestant contestant)
         Id = contestant.Id;
         if (contestant.StartTimeOverride != null)
         {
-            TimeSpan? startTime = contestant.StartTimeOverride.Value.DateTime.TimeOfDay;
-            StartTimeOverride = startTime;
+            StartTimeOverride = contestant.StartTimeOverride.Value.LocalDateTime.TimeOfDay;
         }
```
achobanov, line 22: Much better, but I think it can still be simplified:
```
StartTimeOverride = contestant.StartTimeOverride?.LocalDateTime.TimeOfDay
```

src-v2/Judge/NTS.Judge.Blazor/Pages/Setup/Events/EventPage.razor
```diff
@@ -35,7 +35,7 @@
         var entity = await _behind.Read(default);
         if (entity == null)
         {
-           GuardHelper.ThrowIfNull(entity, "Event cannot be null here");    
+           GuardHelper.ThrowIfNull(entity);    
         }
```
achobanov, line 39: Haha, not quite. Take a careful look at lines I've marked here and then look at the definition of `ThrowIfNull` again. If you still can't see it - replace the invocation of `ThrowIfNull`  with it's definition and you'll see what I mean.

src-v2/NTS.Domain.Setup/Entities/Contestant.cs
```diff
@@ -20,18 +20,18 @@ private Contestant(int id, DateTimeOffset? startTimeOverride, bool unranked) : t
...
     private Contestant(DateTimeOffset? startTimeOverride, bool unranked)
     {
-        if (startTimeOverride!=null && startTimeOverride.Value.DateTime.CompareTo(DateTime.Today) < 0)
+        if ( startTimeOverride != null && startTimeOverride.Value.DateTime.CompareTo(DateTime.Today) < 0)
         {
             throw new DomainException(nameof(StartTimeOverride), "Start time cannot be in the past");
         }
         StartTimeOverride = startTimeOverride;
-        Unranked = unranked;
+        IsUnranked = unranked;
```
achobanov, line 28: Rename variable name as well

src-v2/Not/Not.Blazor/TM/NotChildren.razor
```diff
@@ -45,18 +47,25 @@ else
     public string NoItemsMessage { get; set; } = default!;
     [Parameter]
     public string UpdateRoute { get; set; } = default!;
+    [Parameter]
+    public string Class { get; set; } = default!;
```
achobanov, line 51: Do you think there is value in exposing parameters that aren't being used? It's not a hard thing to add a parameter if you need it at some point. And allowing unnecessary configuration can lead to overengineering and inconsistencies, both of which aren't ideal. Same goes for the `NoItemsMessage`.

src-v2/Not/Not.Blazor/TM/NotChildren.razor
```diff
@@ -45,18 +47,25 @@ else
...
 
     public IEnumerable<T> Items { get; set; } = default!;
 
     protected override void OnInitialized()
     {
+        Name = Localizer.Get(Name);
+        if(NoItemsMessage == default)
```
achobanov, line 58: I would rename to the simpler `EmptyMessage`.

src-v2/Not/Not.Blazor/TM/NotChildren.razor
```diff
@@ -45,18 +47,25 @@ else
...
+        {
+            string defaultMessage = $"No {Name.ToLower()} have been created for this event.";
+            NoItemsMessage = Localizer.Get(defaultMessage);
```
achobanov, line 61: When defined like this you will have to add multiple entries in the localization resex files - one for every usage of `NotChildren` with a different `Name`. Instead you could define the property as a constant that contains a string placeholder and use `string.Format`. That way we will have to translate a single resex entry.  However this isn't currently supported by the `Ilocalizer` so you can just leave a TODO comment so we don't forget about it later.

src-v2/Not/Not/Extensions/DateTimeOffsetExtension.cs
```diff
@@ -13,11 +14,10 @@ public static DateTimeOffset ToDateTimeOffset(this DateTime time)
...
     }
-    public static DateTimeOffset? ToDateTimeOffset(this TimeSpan? timeToBeAdded)
+    public static DateTimeOffset ToDateTimeOffset(this TimeSpan timeToBeAdded)
     {
-        if(timeToBeAdded == null) return null;
         DateTime today = DateTime.Today;
-        var time = today.Add((TimeSpan)timeToBeAdded);
+        DateTime time = today.Add(timeToBeAdded);
         DateTime timeWithSpecifiedKind = DateTime.SpecifyKind(time, DateTimeKind.Local);
         DateTimeOffset offsetTime = timeWithSpecifiedKind;
```
achobanov, line 22: Implicit vars again, It seems you have a habit of using them (or is it the IDE somehow? ). Try to be more vigilant and replace then with `var` whenever you see them, until you make a habit of using it in the first place.

achobanov (changes requested):

achobanov (changes requested):
Looks good, but I've found a bug with Contestant Update - StartTimeOverride is kept, but  it's checkbox is always off. which isnt correct. Also do you think we should add `StartTimeOverride` and `IsUnranked` in ToString?

achobanov (approved):

PR #221, opened 2024-05-08
NTS/Refactor loop entity in setup boundary

StoyanGoranov:
Closes: https://github.com/Not-Endurance/not-timing-system/issues/210


src-v2/NTS.Domain.Setup/Entities/Contestant.cs
```diff
@@ -42,7 +42,7 @@ public override string ToString()
     {
         var sb = new StringBuilder();
         var startTimeMessage = StartTimeOverride == null ? "" : "StartTime: " + $"{StartTimeOverride.Value.ToLocalTime().TimeOfDay}";
-        var IsUnrankedMessage = IsUnranked ? "Unranked: Yes" : "Unranked: No";
+        var IsUnrankedMessage = IsUnranked ? "Unranked" : "Ranked";
```
achobanov, line 45: `isUnrankedMessage`

src-v2/NTS.Domain.Setup/Entities/Contestant.cs
```diff
@@ -42,7 +42,7 @@ public override string ToString()
...
         sb.Append($"Contestant -> Tandem: *isn't configured yet* {startTimeMessage} {IsUnrankedMessage} ");
```
achobanov, line 46: If `startTimeMessage`  is empty here it would lead to duplicate spaces, right? Isn't there a better solution?

src-v2/NTS.Domain.Setup/Entities/Lap.cs
```diff
@@ -0,0 +1,36 @@
...
+        }
+
+        Distance = distance;
+    }
+    public double Distance { get; set; }
+
+    public string NestedToString()
+    {
+        return Distance.ToString()+"km";
+    }
```
achobanov, line 28: NestedToString? Nested how? Is this ncessary?

src-v2/NTS.Domain.Setup/Entities/Lap.cs
```diff
@@ -0,0 +1,36 @@
...
+    public override string ToString() 
+    {
+        var phase = "Lap".Localize();
+        var sb = new StringBuilder();
+        sb.Append($"{phase} -> {Distance}km long ");
+        return sb.ToString();
```
achobanov, line 34: Seems a bit overcomplicated. Do you have yo use StringBuilder here?

src-v2/NTS.Domain.Setup/Entities/Loop.cs
```diff
@@ -1,51 +1,58 @@
...
+    public static Loop Create(double distance) => new(distance);
+    public static Loop Update(int id, double distance) => new(id, distance);
 
+    public Loop()
+    {
+        Laps = new List<Lap> { };
+    }
     [JsonConstructor]
-    public Loop(int id, double distance, int recovery, int rest) : this(distance, recovery, rest)
+    public Loop(List<Lap> laps)
```
achobanov, line 14: Wait, Loop consists of Laps? Now I am entirely confused. I don't think we need all 3 of `Phase`, `Lap` and `Loop`.

src-v2/NTS.Domain.Setup/Entities/Loop.cs
```diff
@@ -1,51 +1,58 @@
...
+        var newLap = new Lap(distance);
+        Laps.Add(newLap);
+    }
+    public Loop(int id, double distance) 
     {
         Id = id;
+        Laps.RemoveAll(lap => lap.Id == id);
+        var updatedLap = new Lap(distance);
+        Laps.Add(updatedLap);
+    }
```
achobanov, line 33: This constructor in particular seems wrong.

src-v2/NTS.Domain.Setup/Entities/Phase.cs
```diff
@@ -0,0 +1,52 @@
...
+        }
+        if (rest <= 0)
+        {
+            throw new DomainException(nameof(Rest), "Rest duration cannot be zero or less.");
+        }
+        SelectedLap = selectedLap;
+        Recovery = recovery;
+		Rest = rest;
+	}
+    public Lap? SelectedLap { get; private set; }
```
achobanov, line 33: Do you think **Selected** adds meaningful context here?

src-v2/NTS.Domain.Setup/Entities/Phase.cs
```diff
@@ -0,0 +1,52 @@
...
+    public int Recovery { get; private set; }
+    public int Rest { get; private set; }
+
+    public void SetSelectedLap(Lap selectedLap)
+    {
+        SelectedLap = selectedLap;
+    }
```
achobanov, line 40: Method setters like this should be avoidable. You could simply make the `SelectedLap` property with a public or internal setter. 

Furthermore, are you sure this methods works correctly? Have you  tried to update a phase after it's Lap has been set?

src-v2/NTS.Domain.Setup/Entities/Phase.cs
```diff
@@ -0,0 +1,52 @@
...
+    public override string ToString()
+    {
+        var min = "min".Localize();
+		var rec = "Recovery".Localize();
+        var lap = "Lap".Localize();
+        var phase = "Phase".Localize();
+		var rest = "Rest".Localize();
+        var selectedLap = SelectedLap==null? "DELETED".Localize() : SelectedLap.NestedToString();
```
achobanov, line 48: Space around operators.

src-v2/NTS.Persistance/Adapters/SetupLapRepository.cs
```diff
@@ -0,0 +1,44 @@
+﻿using NTS.Domain.Setup.Entities;
+using NTS.Persistence.Setup;
+
+namespace NTS.Persistence.Adapters;
+
+public class SetupLapRepository : SetRepository<Lap, SetupState>
+{
+    public SetupLapRepository(IStore<SetupState> store) : base(store)
```
achobanov, line 8: I don't think the `Setup` prefix is necessary here unless we decide to prefix all our repositories. Currently I'm only prefixing duplicate names and there isn't a `Lap` nor `Loop` in Core.

src-v2/NTS.Persistance/Adapters/SetupLapRepository.cs
```diff
@@ -0,0 +1,44 @@
...
+    {
+        state = store.Load().Result;
+    }
+
+    private SetupState state { get; set; }
```
achobanov, line 13: Use private fields instead of properties. By MS convention a field should only ever be private, while properties can be protected or public.

src-v2/NTS.Persistance/Adapters/SetupLapRepository.cs
```diff
@@ -0,0 +1,44 @@
+﻿using NTS.Domain.Setup.Entities;
+using NTS.Persistence.Setup;
+
+namespace NTS.Persistence.Adapters;
+
+public class SetupLapRepository : SetRepository<Lap, SetupState>
+{
+    public SetupLapRepository(IStore<SetupState> store) : base(store)
+    {
+        state = store.Load().Result;
```
achobanov, line 10: Perhaps I didn't emphasize enough that we don't want `Task.Result` in the codebase. At best it negates all benefits from asynchronous code, at worst it can halt the app.

Furthermore, constructors **should not** contain asynchronous logic. It's especially dangerous to call `.Result` here, because this is a **Dependency**, i.e. a service managed by the Dependency Injection Container. The container uses this constructor in ways you do not undestand and very often they are used during app startup, when configurations have not been executed yet. If this is a call to an actual database it will probably throw connectivity-related exceptions. Constructors are meant for a lightweight object initialization that can be used by the C# runtime on demand. **It's not a coincidence that they do not support async syntax**

src-v2/NTS.Persistance/Adapters/SetupLapRepository.cs
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    private SetupState state { get; set; }
+    public override Task<Lap> Update(Lap entity)
```
achobanov, line 14: As a general rule of thumb always use `async` with methods that return Task. [A good explanation why here](https://blog.stephencleary.com/2016/12/eliding-async-await.html)

src-v2/NTS.Persistance/Adapters/SetupLapRepository.cs
```diff
@@ -0,0 +1,44 @@
...
+    {
+        state.Event.Competitions
+            .Where(competition => competition.Phases.Any(phase => phase.SelectedLap.Id == entity.Id))
```
achobanov, line 17: You can use `phase.SelectedLap == entity` due to `Equals` and `GetHashCode` override sin `DomainEntity.

src-v2/NTS.Persistance/Adapters/SetupLapRepository.cs
```diff
@@ -0,0 +1,44 @@
...
+            .SelectMany(competition => competition.Phases
+                .Where(phase => phase.SelectedLap.Id == entity.Id)
+                .Select(phase => new { competition, phase }))
+            .ToList()
+            .ForEach(item =>
+            {
+                item.phase.SetSelectedLap(entity);
+                item.competition.Update(item.phase);
+            });
```
achobanov, line 26: Isn't
```c#
foreach (var phase in Event.Competitions.SelectMany(x => x.Phases))
{
    if (phase == entity)
    {
        phase.SetSelectedLap(entity);
    }
}
```
Simpler?

Additionally, the first `Where` clause is redundant as you are making the same check again in `SelectMany`.  Also `ToList` isn't necessary.

src-v2/NTS.Persistance/Adapters/SetupLapRepository.cs
```diff
@@ -0,0 +1,44 @@
...
+            .Where(competition => competition.Phases.Any(phase => phase.SelectedLap.Id == entity.Id))
+            .SelectMany(competition => competition.Phases
+                .Where(phase => phase.SelectedLap.Id == entity.Id)
+                .Select(phase => new { competition, phase }))
+            .ToList()
+            .ForEach(item =>
+            {
+                item.phase.SetSelectedLap(null);
+                item.competition.Update(item.phase);
+            });
```
achobanov, line 41: Seems like the same code as in `Update`. Can you not reuse it?

src-v2/NTS.Persistance/States/SetupState.cs
```diff
@@ -1,14 +1,20 @@
...
+
     Event? ITreeState<Event>.Root
     {
         get => Event; 
         set => Event = value;
     }
+
+    List<Lap> ISetState<Lap>.EntitySet { 
+        get => Laps;
+    }
```
achobanov, line 19: `List<Lap> ISetState<Lap>.EntitySet => Laps;` is enough

src-v2/Not/Not.Application/Ports/CRUD/ISetRepository.cs
```diff
@@ -0,0 +1,8 @@
+﻿using Not.Domain;
+
+namespace Not.Application.Ports.CRUD;
+
+public interface ISetRepository<T> : ICreate<T>, IReadAll<T>, IUpdate<T>, IDelete<T>
```
achobanov, line 5: Good job here. This shows a level of understanding of composition of our `IRepository` abstractions.

src-v2/Not/Not.Blazor/TM/NotChildren.razor
```diff
@@ -11,20 +11,20 @@
...
     {
-        <p class="mt-1 ml-6 mb-1">
-            @item
-            <div class="action-column">
-                <NotButtonUpdate Size="Size.Small" OnClick="() => this._contestantNavigator.Update(UpdateRoute, item)" />
-                <NotButtonDelete Size="Size.Small" OnClick="() => _parentBehind.Delete(item)" />
+        <div class="d-grid mt-1 ml-6 mb-1">
+            <div class="grid-row-1">@item</div>
+            <div class="grid-row-1 action-column">
+                <NotButtonUpdate  Size="Size.Small" OnClick="() => this._contestantNavigator.Update(UpdateRoute, item)">Edit</NotButtonUpdate>
```
achobanov, line 24: Edit

src-v2/Not/Not.Blazor/TM/NotSetList.razor
achobanov, line 1: This whole file seems like a lot of copy-paste. Does it have to be a separate component if they are so identical?

src-v2/Not/Not.Storage/Stores/InMemoryJsonStore.cs
```diff
@@ -5,7 +5,7 @@ namespace Not.Storage.Stores;
 public class InMemoryJsonStore<T> : IStore<T>
     where T : class, new()
 {
-    private T? _context;
+    private T _context = new();
```
achobanov, line 8: Is `_context` actually used?

src-v2/Not/Not.Storage/Ports/ISetState.cs
achobanov, line 1: Duplicate file caused by merge

achobanov (changes requested):
Overall pretty good, except that there are numerous build errors atm. It's a good habit to always build before pushing changes.

src-v2/Judge/NTS.Judge.Blazor/Pages/Setup/Competitions/CompetitionUpdate.razor
```diff
@@ -11,7 +11,7 @@
 @inherits NotComponent
 
 <MudPaper Elevation="0" Class="pl-6 pr-6">
-    <NotH2>@Localizer.Get("Update Competition")</NotH2>
+    <NotH2>@Localizer.Get("Edit Competition")</NotH2>
```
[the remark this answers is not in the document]
StoyanGoranov: Because from user POV they are about to edit the records, then after editing is completed the Update action is called which isn't performed by the user. My idea is to differentiate these and place the 'responsibility' for editing on the user and the responsibility for updating on the system.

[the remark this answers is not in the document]
achobanov: While probably more semantically correct, I don't find it a meaningful distinction. Instead I prefer to stick to simpler scenario where we have one term and we share terminology with the user, rather then diverge from it. It would be different if from a software standpoint we need to segregate different process stages and call one Edit and another Update, but I don't think that's necessary and desirable here.

src-v2/NTS.Domain.Setup/Entities/Lap.cs
StoyanGoranov: Nested because when Lap is a property of Phase the ToString() method of Phase would need to print only the Distance property of Lap. Alternatively, the Distance property can be called directly.

src-v2/Judge/NTS.Judge.Blazor/Pages/Setup/Competitions/CompetitionUpdate.razor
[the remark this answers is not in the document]
StoyanGoranov: It's not that one way is more semantically correct, but from UX perspective the application should be intuitive to stimulate people to use it. UI should hint to the expected behavior.

 For this reason I haven't changed the labels of the Form buttons as the difference in software behavior is:
- Edit buttons open the *UpdateForm* 
- Update buttons send the edited data for persistence

In that sense I find that from software standpoint we already perform different actions with the different buttons and we call the entire process **Update** although it implicitly include Editing and it is up to the user to initiate the actual update or cancel the changes.

I just think having the same text will be detrimental from UX point of view, however I don't mean to argue as the final decision is yours.

src-v2/Judge/NTS.Judge.Blazor/Pages/Setup/Phases/PhaseForm.razor
```diff
@@ -0,0 +1,57 @@
+﻿@using NTS.Domain.Setup.Entities;
+@using NTS.Judge.Setup.Competitions
+@using Not.Blazor.Forms;
+@using Not.Blazor.TM.Forms;
+@using Not.Blazor.TM.Forms.Components;
+
+@inherits NotForm<Phase>;
+
+<EditForm Model="_model">
+    <MudSelect @ref="_lapField" @bind-Value="_model.SelectedLap" Label="@Localizer.Get("Phase Distance")" @onchange="(_) => SubmitUpdate()"
```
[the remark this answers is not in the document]
StoyanGoranov: Seems like a good idea since MudSelect is already used 3 times but there is a problem with the implementation since each MudSelect may represent a selectable list of different object types.

However the NotSelect must accept a generic type parameter:
`@typeparam T where T : DomainEntity`

An issue arises immediately at the `OfficialForm.razor` component because it uses and Enum to provide the items of the Select.

src-v2/Judge/NTS.Judge.Blazor/Pages/Setup/Competitions/CompetitionUpdate.razor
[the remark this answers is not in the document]
achobanov: I see your point, but honestly I don't think people would make that distinction. I know I didn't. Also not entirely sure it's correct: [quick google search](https://www.google.com/search?q=what%27s+the+difference+between+edit+and+update&rlz=1C1YTUH_enBG1091BG1091&oq=what%27s+the+difference+between+edit+and+update&gs_lcrp=EgZjaHJvbWUyBggAEEUYOTIHCAEQABiABDIHCAIQABiABDIHCAMQABiABDIHCAQQABiABDIHCAUQABiABDIHCAYQABiABDIHCAcQABiABDIHCAgQABiABDIHCAkQABiABNIBCDUyNDdqMGo3qAIAsAIA&sourceid=chrome&ie=UTF-8). 

I think there are better ways to spend our time. Lets keep Update everywhere and for the future lets discuss things that fall outside of the scope of the issue, before implementing them and then having to revert back later.

src-v2/NTS.Domain.Setup/Entities/Lap.cs
achobanov: That means that you are defining a member of Lap with the added context of Phase as it's parent. That's wrong. Lap doesn't know and shouldn't know how it is used. This is true for all objects, not only Domain ones.

ToString in Domain entities should be agnostic of their usage - i.e. they should provide relevant but not excessive information. In the example bellow "long" is completely unnecessary and the "Lap -" is questionable

achobanov:
Also a reminder for the branch name convention - nts/<feature> or dev/<feature> (if working on the legacy app)

src-v2/Judge/NTS.Judge.Blazor/Pages/Setup/Competitions/CompetitionForm.razor
```diff
@@ -45,7 +42,7 @@
     {
         RegisterInjector(nameof(_model.Name), () => _nameField);
         RegisterInjector(nameof(_model.StartTime), () => _timeField);
-        RegisterInjector(nameof(_model.Type), () => _typeField);
+        //RegisterInjector(nameof(_model.Type), () => _typeField);
```
achobanov, line 45: Is there a problem with the validation?

src-v2/Judge/NTS.Judge.Blazor/Pages/Setup/Officials/OfficialForm.razor
```diff
@@ -4,29 +4,35 @@
...
-    private MudSelect<OfficialRole> _roleField = default!;
+    private NotSelect<OfficialRole> _roleField = default!;
     private OfficialFormModel _model { get; set; } = new();
 
+    private IEnumerable<NotListModel<OfficialRole>> _items = NotListModel.FromEnum<OfficialRole>();
+
+    protected override void OnInitialized()
+    {
+        base.OnInitialized();
+    }
```
achobanov, line 29: Seems unnecessary

src-v2/Judge/NTS.Judge.Blazor/Pages/Setup/Officials/OfficialForm.razor
```diff
@@ -4,29 +4,35 @@
...
+
     public override void RegisterValidationInjectors()
     {
         RegisterInjector(nameof(_model.Name), () => _nameField);
-        RegisterInjector(nameof(_model.Role), () => _roleField);
-    }
+        // RegisterInjector(nameof(_model.Role), () => _roleField);
```
achobanov, line 34: Validation?

src-v2/Judge/NTS.Judge.Blazor/Pages/Setup/Phases/PhaseForm.razor
```diff
@@ -1,41 +1,40 @@
...
 
 @code {
-    private MudSelect<Lap> _lapField = default!;
+    private NotSelect<Loop> _loopField = default!;
     private MudNumericField<int> _recoveryField = default!;
     private MudNumericField<int> _restField = default!;
     private PhaseFormModel _model { get; set; } = new();
-    private IEnumerable<Lap> Items { get; set; } = default!;
+
+    private IEnumerable<NotListModel<Loop>> _items = default!;
```
achobanov, line 24: Isn't `_items` a bit generic? What happens if we add a second `NotSelect` for example?

src-v2/Judge/NTS.Judge.Blazor/Pages/Setup/Phases/PhaseForm.razor
```diff
@@ -1,41 +1,40 @@
...
     {
-        Items = await _setBehind.GetAll();
+        var items = await _setBehind.GetAll();
+        _items = NotListModel.FromEntity<Loop>(items);
     }
 
     public override void RegisterValidationInjectors()
     {
-        RegisterInjector(nameof(_model.SelectedLap), () => _lapField);
+        //RegisterInjector(nameof(_model.Loop), () => _loopField);
```
achobanov, line 37: Validation

src-v2/Judge/NTS.Judge/Events/CompetitionChildrenBehind.cs
```diff
@@ -2,9 +2,10 @@
 using Not.Blazor.Ports.Behinds;
 using Not.Exceptions;
 using NTS.Domain.Setup.Entities;
+using System.Collections.Generic;
 
 namespace NTS.Judge.Events;
-public class CompetitionChildrenBehind : INotBehindParent<Contestant>, INotBehindParent<Phase>, INotBehindWithChildren<Competition>
+public class CompetitionChildrenBehind : INotSetBehind<Contestant>, INotSetBehind<Phase>, INotBehindWithChildren<Competition>
```
achobanov, line 8: This might be a good place to rename the odd `INotBehindWithChildren` to a more reasonable `INotBehindParent` since  we no longer have another interface conflicting.

src-v2/NTS.Domain.Setup/Entities/Loop.cs
```diff
@@ -1,58 +1,30 @@
...
-        Laps.Remove(child);
-    }
-
-    public void Update(Lap child)
+    public override string ToString() 
     {
-        Laps.RemoveAll(lap => lap.Id == child.Id);
-        Laps.Add(child);
+        var message = $"{Distance}km";
+        return message;
```
achobanov, line 28: Can be single line. Also km isn't localized.

src-v2/NTS.Domain.Setup/Entities/Phase.cs
```diff
@@ -4,19 +4,19 @@
...
 {
-    public static Phase Create(Lap lap, int recovery, int rest) => new(lap, recovery, rest);
+    public static Phase Create(Loop Loop, int recovery, int rest) => new(Loop, recovery, rest);
 
-    public static Phase Update(int id, Lap lap, int recovery, int rest) => new(id, lap, recovery, rest);
+    public static Phase Update(int id, Loop Loop, int recovery, int rest) => new(id, Loop, recovery, rest);
 
     [JsonConstructor]
-    public Phase(int id, Lap selectedLap, int recovery, int rest)
+    public Phase(int id, Loop selectedLoop, int recovery, int rest)
```
achobanov, line 12: Isn't this `selectedLoop`  name a problem? Does `selected` bring meaningful context?

src-v2/NTS.Persistance/Adapters/LoopRepository.cs
```diff
@@ -3,54 +3,54 @@
...
-    public override async Task<Lap> Delete(Lap entity)
+    public override async Task<Loop> Delete(Loop entity)
     {
-        await UpdateLapValue(entity,true);
+        await UpdateLoopValue(entity,true);
         return entity;
     }
 
-    public async Task UpdateLapValue(Lap entity, bool isDelete=false) 
+    public async Task UpdateLoopValue(Loop entity, bool isDelete=false)
```
achobanov, line 26: Does it have to be public?

achobanov (changes requested):

src-v2/Judge/NTS.Judge.Blazor/Pages/Setup/Competitions/CompetitionForm.razor
StoyanGoranov: Currently, the RegisterInjector only accepts `Func<MudBaseInput<TInput>>` or `Func<MudPicker<TInput>>` as a second argument. We need to extend it to also support Custom components because the _typeField is no longer of type `MudSelect<CompetitionType>` but `NotSelect<CompetitionType>` instead.

I have left a comment:
// TODO: Implement Validation injection for custom input components
at line 17 NotSelect.razor to ensure the needed improvements aren't forgotten.

src-v2/Judge/NTS.Judge/Events/CompetitionChildrenBehind.cs
StoyanGoranov: Renamed the interface, however used the name `INotParentBehind` to avoid the need to remove the `INotBehindParent` yet. And also comply with the emerging naming convention -> INot-*semantic part of the name*-Behind. 

Looking forward to hearing what you think about that approach.

achobanov: Sounds great, I like consistency. However why not delete the old `INotBehindParent`?

src-v2/Judge/NTS.Judge.Blazor/Pages/Setup/Competitions/CompetitionForm.razor
achobanov: Makes sense. It's probably going to stay like that for the demo version and then for NTS 1 we'll probably end up reworking the validation anyway.

achobanov (approved):
Good job on renaming `_items` in multiple places, not only where I had left a comment

src-v2/Judge/NTS.Judge/Events/CompetitionChildrenBehind.cs
StoyanGoranov: Done.

PR #244, opened 2024-06-18
NTS/Fix competition bug

StoyanGoranov:
This is needed to enable the children of parent entities to be updated when their respective sets are modified.

### DIFF
--- a/src-v2/Judge/NTS.Judge.Blazor/Pages/Setup/Competitions/CompetitionUpdate.razor
+++ b/src-v2/Judge/NTS.Judge.Blazor/Pages/Setup/Competitions/CompetitionUpdate.razor
@@ -21,7 +21,7 @@
     <NotSetList Name="Contestants" T="Contestant" TForm="ContestantForm" UpdateRoute="@CONTESTANT_UPDATE" />
 </MudPaper>
 
-@code {   
+@code {
     private Competition _competition { get; set; } = default!;
     [Inject]
     private INotSetBehind<Competition> _setBehind { get; set; } = default!;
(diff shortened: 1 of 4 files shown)


src-v2/Judge/NTS.Judge.Blazor/Pages/Setup/Competitions/CompetitionUpdate.razor
```diff
@@ -32,12 +32,13 @@
 
     protected override async Task OnInitializedAsync()
     {
-        _competition = _navigator.ConsumeParameter<Competition>();
-        await _competitionWithChildren.Initialize(_competition.Id);
+        var competition = _navigator.ConsumeParameter<Competition>();
+        _competition = await _competitionWithChildren.Initialize(competition.Id);
```
achobanov, line 36: I don't understand. what's the difference between having `_competition` be assigned a reference from `ConsumeParameter` or from `Initialize`?

achobanov (commented):
Not sure this is necessary. If it is then how does it work in `EventUpdate.razor`? I don't see a change there.

Furthermore, I've updated the PR name to better represent what problem it solves.

src-v2/Judge/NTS.Judge.Blazor/Pages/Setup/Competitions/CompetitionUpdate.razor
StoyanGoranov: Upon debugging I have found that the data loss behavior arises because the object returned by the ConsumeParameter at CompetitionUpdate isn't related to its children. I. e the Contestants and Phases fields are empty. When this state is persisted the data loss occurs.

However, with the object returned by the Initialize method at the INotParentBehind the fields are populated as they have been stored in-memory. 

Why the INavigator works as expected at Event but doesn't at Competition is also unclear to me.

achobanov (approved):

PR #270, opened 2024-08-13
nts/core Actions panel

### LINKED ISSUE #251: Create Actions panel - ReInspection, Required Inpsection, FTQ/DQ/RET

StoyanGoranov:
Closes: https://github.com/Not-Endurance/not-timing-system/issues/251


src-v2/Judge/NTS.Judge/Adapters/Behinds/ParticipationBehind.cs
```diff
@@ -24,24 +25,46 @@ public ParticipationBehind(
...
+        }
+        GuardHelper.ThrowIfDefault(SelectedParticipation);
+
+        var result = SelectedParticipation?.Process(snapshot);
         if (result.Type == SnapshotResultType.Applied)
         {
-            await _participationRepository.Update(participation);
+            await _participationRepository.Update(SelectedParticipation);
         }
         await _snapshotResultRepository.Create(result);
```
achobanov, line 69: Revert changes in this method as discussed.

src-v2/Judge/NTS.Judge/Adapters/Behinds/ParticipationBehind.cs
```diff
@@ -50,18 +73,24 @@ public async Task Process(Snapshot snapshot)
...
 
         participation.Update(state);
         await _participationRepository.Update(participation);
     }
 
-    public async Task RevokeQualification(int number, QualificationRevokeType type, FTQCodes? ftqCode = null, string? reason = null)
+    public async Task Update(Participation participation)
+    {
+        await _participationRepository.Update(participation);
+    }
```
achobanov, line 87: Is this method necessary?

src-v2/Judge/NTS.Judge/Adapters/Behinds/ParticipationBehind.cs
achobanov: Actually, revert the changes in the whole file then we'll fix it using ObservableBehind after we merge all branches.

src-v2/NTS.Domain.Core/Aggregates/Participations/NotQualified.cs
```diff
@@ -123,4 +158,5 @@ protected NotQualified(string complement)
     }
 
     public string? Complement { get; private set; }
+    public virtual string Type { get; set; }
```
achobanov, line 161: Remove as discussed

src-v2/NTS.Domain.Core/Aggregates/Participations/NotQualified.cs
```diff
@@ -51,24 +63,38 @@ public record FailedToQualify : NotQualified
...
             throw new DomainException($"'Failed to Complete' requires a writen explanation from officials. Please provide 'complement'");
         }
-        Code = code;
+        Codes = codes;
     }
     public FailedToQualify(string complement) : base(complement)
     {
-        Code = FTQCodes.FTC;
+        Type = nameof(FailedToQualify);
+        Codes.ToList().Add(FTQCodes.FTC);
```
achobanov, line 78: This is bugged as discussed.

src-v2/NTS.Domain.Core/Aggregates/Participations/NotQualified.cs
```diff
@@ -51,24 +63,38 @@ public record FailedToQualify : NotQualified
...
     }
 
-    public FTQCodes Code { get; private set; }
+    public IEnumerable<FTQCodes> Codes { get; private set; }
```
achobanov, line 81: Always initialize collection fields to an empty collection.

src-v2/NTS.Domain.Core/Aggregates/Participations/NotQualified.cs
```diff
@@ -51,24 +63,38 @@ public record FailedToQualify : NotQualified
...
+            if (i < Codes.Count() - 1)
+            {
+                codes += Codes.ToList()[i].ToString() + "+";
+            }
+            else
+            {
+                codes += Codes.ToList()[i].ToString();
+            }      
+        }
+        return $"FTQ {codes}";
```
achobanov, line 97: Can use `string.Join` instead.

src-v2/NTS.Domain.Core/Aggregates/Participations/Participation.cs
```diff
@@ -70,7 +74,18 @@ public void Update(IPhaseState state)
         phase.Update(state);
         EvaluatePhase(phase);
     }
-
+    public void IsReinspectionRequested(bool requested)
+    {
+        GuardHelper.ThrowIfDefault(Phases);
```
achobanov, line 79: Phases is never null.

src-v2/NTS.Domain.Core/Aggregates/Participations/Participation.cs
```diff
@@ -70,7 +74,18 @@ public void Update(IPhaseState state)
...
+        GuardHelper.ThrowIfDefault(Phases.Current);
```
achobanov, line 80: Current can be null, but this is not a good place for a GuardException. Any idea why?

src-v2/NTS.Domain.Core/Aggregates/Participations/Participation.cs
```diff
@@ -70,7 +74,18 @@ public void Update(IPhaseState state)
         phase.Update(state);
         EvaluatePhase(phase);
     }
-
+    public void IsReinspectionRequested(bool requested)
```
achobanov, line 77: `IsReinspectionRequested` seems semantically wrong to me. The name of the method poses a question and simultaneously sets state. I would expect method with such name to return a bool result, not use one as an argument. Methods that set values should read like commands.

src-v2/NTS.Domain/Objects/Timestamp.cs
```diff
@@ -28,6 +28,11 @@ public Timestamp Add(TimeSpan span)
         return new Timestamp(DateTime.Add(span));
     }
 
+    public Timestamp SetTime(TimeSpan span)
+    {
+        return new Timestamp(System.DateTime.Today.Add(span));
+    }
```
achobanov, line 34: I don't like this method. The main idea of the Timestamp is to represent a snapshot in the current time. It doesn't make much sense to Set it's time. But there is a bigger problem. Any idea what it might be?

src-v2/Not/Not.Blazor/TM/Models/MultiSelectListModel.cs
```diff
@@ -0,0 +1,36 @@
+﻿using System.ComponentModel;
+using System.Linq.Expressions;
+using System.Reflection;
+
+namespace Not.Blazor.TM.Models;
```
achobanov, line 5: Does this class has anything to do with Blazor?

src-v2/Not/Not.Blazor/TM/Models/MultiSelectListModel.cs
```diff
@@ -0,0 +1,36 @@
...
+
+public class MultiSelectListModel
```
achobanov, line 7: Does this class has anything to do with multi-select? It seems to represent just a pair of Value and Description

src-v2/Not/Not.Blazor/TM/Models/MultiSelectListModel.cs
```diff
@@ -0,0 +1,36 @@
...
+        Description = description;
+    }
+
+    public string Value { get; set; }
+    public string Description { get; set; }
+
+    public static IEnumerable<MultiSelectListModel> MultiSelectList<T>()
+    {   
+        var enumValues = Enum.GetValues(typeof(T)).Cast<Enum>();
+        var selectItems = enumValues.Select(s => new MultiSelectListModel(s.ToString(), EnumHelper.GetDescription(s)));
```
achobanov, line 21: `EnumHelper.GetDescription(s)`  can be  `s.GetDescription()`, because GetDescription is defined as an extension method.

src-v2/Not/Not.Blazor/TM/Models/MultiSelectListModel.cs
```diff
@@ -0,0 +1,36 @@
...
+        return selectItems;
+    }
+}
+
+public static class EnumHelper
```
achobanov, line 26: Does it have anything to do with Blazor as well? Is it strictly specific for `MultiSelectListModel` only?

src-v2/Not/Not.Blazor/TM/Models/MultiSelectListModel.cs
```diff
@@ -0,0 +1,36 @@
...
+{
+    public static string GetDescription(this Enum value)
+    {
+        FieldInfo fieldInfo = value.GetType().GetField(value.ToString());
+        DescriptionAttribute descriptionAttribute = fieldInfo.GetCustomAttributes(typeof(DescriptionAttribute), false)
+                                                             .FirstOrDefault() as DescriptionAttribute;
+        return descriptionAttribute == null ? value.ToString() : descriptionAttribute.Description;
```
achobanov, line 33: Overall pretty good. I have a few formatting grapes, but more importantly: keep in mind that when we are using reflection we want to be doing that using the `ReflectionHelper` class.

src-v2/Judge/NTS.Judge.MAUI/wwwroot/css/app.css
```diff
@@ -34,4 +43,31 @@ th.mud-table-cell {
...
+    background-color: rgb(19,6,135) !important;
+    border-color: black !important;
+    border-width: 0.5px !important;
+}
+
+.secondary-outlined-toggled {
+    color: white !important;
+    background-color: rgb(244, 67, 54) !important;
+    border-color: rgb(244, 67, 54) !important;
+    border-width: 0.5px !important;
```
achobanov, line 72: Not a fan of custom css, but we can leave that for v1.

achobanov (changes requested):

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Components/ActionsPanelComponent.razor
```diff
@@ -0,0 +1,334 @@
...
+    public int TandemNumber { get; set; }
+    private MudTextField<string> timeInput = default!;
+    private MudBaseButton reinspectionButton = default!;
+    private MudButton reqInsButton = default!;
+    private MudTextField<string> reasonInput = default!;
+    private MudSelect<string> multiSelect = default!;
+    private Participation? _selectedParticipation = default!;
+    private Phase? _currentPhase = default!;
+    private Snapshot _snapshot = default!;
+    private string Time { get; set; } = default!;
```
achobanov, line 68: Time has to be nullable. You should be able to update without setting Arrive or Inspection times.

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Components/ActionsPanel.razor
```diff
@@ -0,0 +1,298 @@
...
+    private string MultiSelectClasses = "hidden";
+    private string Reason = "";
+    private string FTQReasonValue = "";
+
+    protected override async Task OnInitializedAsync()
+    {
+        await Observe(_participationBehind);
+    }
+
+    public void BeforeRender()
```
achobanov: Encapsulating the logic in BeforeRender method, which you invoke in the render tree is not really a solution. It's better than bare-bones logic among the component, but the logic itself, which is the main problem is still very shaky.  

Why it is shaky:
- By far mostly because it has a large and messy shared scope. A shared scope means it's a scope that contains values reused by multiple components. Naturally it's messy. Reading through this method is very hard as you need to constantly scroll the file in order to see how and where all these shared values are used. 
- Also contains a lot of magic strings. Using value type literals in order to control behavior is sometimes necessary, but should be avoided when possible. Again, having so many does not help. 
- Contains large switch (same applies for `UpdateParticipation` using a large if-else chain) 

We should discuss this in person, but in the meantime think about the following:
- What is the root cause of the problem?
- How can we improve large shared scopes?

Try thinking how you would build this component from the ground up if you had to start again, having had that experience.

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Components/ActionsPanel.razor
```diff
@@ -0,0 +1,298 @@
...
+    }
+
+    private async void ToggleReinspection(bool value)
+    {
+        ReinspectionToggled = value;
+        _participationBehind.SelectedParticipation?.ReinspectionRequested(value);
+        await InvokeAsync(StateHasChanged); // TODO test without as it should re-render on its own
+    }
+
+    private async void ToggleReqIns(bool value)
```
achobanov: Lets start using the full names regarding Domain operations, i.e. RequiredInspection

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Components/ActionsPanel.razor
```diff
@@ -0,0 +1,298 @@
...
+
+        }
+        else if(value == "FailedToQualify")
+        {
+            AddInput(ref MultiSelectClasses);   
+        }
+        ToggleGroupSelection = value;
+    }
+
+    public async Task UpdateParticipant()
```
achobanov: You should always avoid large if-else/switch statements. Think about the similarities between `UpdateParticipant` and `BeforeRender`. What do they have in common?

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Components/ActionsPanel.razor
```diff
@@ -0,0 +1,298 @@
...
+            snapshotType = SnapshotType.Stage;
+        }
+        else if (currentPhase.InspectTime == null && currentPhase.IsFinal == false)
+        {
+            snapshotType = SnapshotType.Vet;
+        }
+        else
+        {
+            snapshotType = SnapshotType.Final;
+        }
```
achobanov: This doesn't seem right.

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Components/DashboardComponent.razor
```diff
@@ -68,8 +61,8 @@
             return;
         }
         _behind.SelectParticipation((int)chip.Value);
-        GuardHelper.ThrowIfDefault(_selectedParticipation);
-        await InvokeAsync(StateHasChanged);
+        GuardHelper.ThrowIfDefault(_behind.SelectedParticipation);
+        await InvokeAsync(StateHasChanged); // TODO: think this sisnt necessary
```
achobanov: Is it necessary?

src-v2/Judge/NTS.Judge.MAUI/wwwroot/css/app.css
```diff
@@ -20,7 +20,7 @@
 }
 
 .hidden{
-    display: none;
+    display: none !important;
 }
```
achobanov: Mud has a [Collapsed](https://mudblazor.com/components/collapse#api) component. Why not use that one?

src-v2/Judge/NTS.Judge/Adapters/Behinds/ParticipationBehind.cs
```diff
@@ -24,32 +23,23 @@ public ParticipationBehind(
 
     public IEnumerable<Participation> Participations { get; private set; } = new List<Participation>();
     public IEnumerable<IGrouping<double, Participation>> ParticipationsByDistance => Participations.GroupBy(x => x.Phases.Distance);
-    public Participation SelectedParticipation { get; private set; } = default!;
+    public Participation? SelectedParticipation { get; set; } = default!;
```
achobanov: Since SelectedParticipation is now nullable you can remove `= default!;`

src-v2/NTS.Domain.Core/Aggregates/Participations/Participation.cs
```diff
@@ -74,15 +73,32 @@ public void Update(IPhaseState state)
...
+        }
+        else if(phase?.InspectTime == null)
+        {
+            return PhaseState.Arrived;
+        }
+        else
+        {
+            return PhaseState.Presented;
+        }
+    }
```
achobanov: This we can ditch that one. As we previously discussed it's an overcomplication that doesn't really matter.

src-v2/NTS.Domain.Core/Aggregates/Participations/Participation.cs
achobanov: The idea is that if `Current` is expected to be null (why would it be defined as nullable then) and you raise a GuardException here, it inevitably is going to be seen by the user. The whole idea of GuardException is NEVER EVER to be seen by the user. They are simply there to help us with the developing process.

achobanov: In this case you can simply remove this, as I'm changing the `Current` property to non-nullable anyway.

src-v2/NTS.Domain.Core/Aggregates/Participations/Phase.cs
```diff
@@ -142,3 +142,10 @@ public interface IPhaseState
     public Timestamp? InspectTime { get; }
     public Timestamp? ReinspectTime { get; }
 }
+
+public enum PhaseState
+{
+    Ongoing = 1,
+    Arrived = 2,
+    Presented = 3
+}
```
achobanov: Not necessary at this phase (:D get it?)

src-v2/Not/Not/Extensions/EnumHelper.cs
```diff
@@ -0,0 +1,16 @@
...
+using Not.Reflection;
+
+namespace Not.Extensions;
+public static class EnumHelper
+{
+    public static string GetDescription(this Enum value)
+    {
+        var type = ReflectionHelper.GetType(value);
+        var field = ReflectionHelper.GetField(type,value);
+        DescriptionAttribute descriptionAttribute = field.GetCustomAttributes(typeof(DescriptionAttribute), false)
```
achobanov: Use `var` instead. Also if we chain method calls we put all on the next line:
```c#
var a = test
    .MethodA()
    .MethodB();
```

src-v2/Not/Not/Reflection/ReflectionHelper.cs
```diff
@@ -19,6 +19,23 @@ public static string GetName<T>()
         return typeof(T).Name;
     }
 
+    public static Type GetType(this object instance)
+    {
+        return instance.GetType();
+    }
```
achobanov: `GetType` is one of the few reflection calls we don't need in `ReflectionHelper`. It's fast enough and there is no real way to optimize it, so simply call `GetType()` on the instance directly.

src-v2/Not/Not/Reflection/ReflectionHelper.cs
```diff
@@ -19,6 +19,23 @@ public static string GetName<T>()
...
+    {
+        if (instance.ToString() != null)
+        {
+            return type.GetField(instance.ToString());
+        }
+        else
+        {
+            return null;
+        }        
+    }
```
achobanov: This isn't a good generic reflection method. Do you have any idea why? What happens if you try to use it on a non-enum instance? How can we improve it?

Also per the Defensive programming paradigm - you should return null at the start and then handle your _success_ path.

src-v2/Not/Not/Reflection/ReflectionHelper.cs
```diff
@@ -36,6 +53,8 @@ public static MethodInfo Method(this Type type, string name)
         return property.GetValue(instance);
     }
 
+
+
```
achobanov: Remove empty lines

src-v2/Not/Not/Structures/SelectListModel.cs
```diff
@@ -0,0 +1,25 @@
...
+    public SelectListModel(string value, string description)
+    {
+        Value = value;
+        Description = description;
+    }
+
+    public string Value { get; set; }
+    public string Description { get; set; }
+
+    public static IEnumerable<SelectListModel> SelectList<T>()
```
achobanov: Lets put static factory methods above the constructor. Additionally, does this name make a lot of semantic sense to you?

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Components/ActionsPanel.razor
```diff
@@ -0,0 +1,298 @@
...
+                    ReasonClasses = "";
+                }
+                break;
+        }
+    }
+
+    private async void ToggleReinspection(bool value)
+    {
+        ReinspectionToggled = value;
+        _participationBehind.SelectedParticipation?.ReinspectionRequested(value);
```
achobanov: Removing the `Is` from `IsReinspectionRequested`  doesn't make the name of the method more commanding. It sounds as if it announces a change, rather than making a change. I.e that's how we name our events. `

achobanov (changes requested):
Still a lot of work here. We should discuss in person.

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Components/ActionsPanel.razor
```diff
@@ -19,14 +19,16 @@
...
-    <MudPaper Class="@("d-flex flex-row mb-4 "+CheckboxDisplay)" Elevation="0">
-        <MudCheckBox T="bool" Value="ReinspectionToggled" ValueChanged="ToggleReinspection" Color="Color.Secondary">
-            <MudIcon Class="pt-3" Style="margin-left: -12px;" Icon="@Icons.Material.Filled._360" V />Reinspection
-        </MudCheckBox>
-        <MudCheckBox T="bool" Value="ReqInsToggled" ValueChanged="ToggleReqIns" Color="Color.Primary">
-            <MudIcon Class="pt-3" Style="margin-left: -12px;" Icon="@Icons.Material.Filled.HourglassEmpty" />Required Inspection
-        </MudCheckBox>
-    </MudPaper>
+    <MudCollapse Expanded="_inspectionActionsExpanded">
+        <MudPaper Class="@("d-flex flex-row mb-4 ")" Elevation="0">
```
achobanov: Unnecessary.

src-v2/Not/Not/Extensions/EnumHelper.cs
```diff
@@ -6,9 +6,8 @@ public static class EnumHelper
 {
     public static string GetDescription(this Enum value)
     {
-        var type = ReflectionHelper.GetType(value);
-        var field = ReflectionHelper.GetField(type,value);
-        DescriptionAttribute descriptionAttribute = field.GetCustomAttributes(typeof(DescriptionAttribute), false)
+        var type = value.GetType();
+        var descriptionAttribute = ReflectionHelper.GetEnumField(type, value)?.GetCustomAttributes(typeof(DescriptionAttribute), false)
                                                              .FirstOrDefault() as DescriptionAttribute;
```
achobanov: formatting

achobanov (changes requested):

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Components/Actions/EliminationForms/DisqualifyForm.razor
```diff
@@ -0,0 +1,48 @@
...
+        set { _value = value; }
+    }
+    private string? _value;
+    private MudTextField<string?> reasonInput = default!;
+
+    public async Task Revoke()
+    {
+        Participation? participation = _participationBehind.SelectedParticipation;
+        var tandemNumber = participation!.Tandem.Number;
+        var reason = Reason;
```
achobanov: Is this variable helpful in any way?

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Components/Actions/EliminationForms/DisqualifyForm.razor
```diff
@@ -0,0 +1,48 @@
...
+    [Parameter]
+    public EventCallback OnSubmission { get; set; }
+
+    private string? Reason 
+    {
+        get => Disqualified != null ? Disqualified.Complement : _value;
+        set { _value = value; }
+    }
+    private string? _value;
+    private MudTextField<string?> reasonInput = default!;
```
achobanov: Is this necessary?

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Components/Actions/EliminationForms/FailedToQualifyForm.razor
```diff
@@ -0,0 +1,102 @@
...
+            @failToQualifyCode.Value
+            <div class="font-size-10">@failToQualifyCode.Description</div>
+        </MudSelectItem>
+    }
+</MudSelect>
+<MudCollapse Expanded="_reasonInputActive">
+    <MudTextField @ref="reasonInput" @bind-Value="Reason" Label="Reason" Variant="Variant.Outlined"></MudTextField>
+</MudCollapse>
+
+
```
achobanov: double empty line

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Components/Actions/EliminationForms/FailedToQualifyForm.razor
```diff
@@ -0,0 +1,102 @@
...
+    private string? _value;
+    public string? Reason 
+    { 
+        get => FailedToQualify != null && FailedToQualify.Codes != null ? FailedToQualify.Complement : _reasonValue;
+        set { _reasonValue = value; }
+    }
+    private string? _reasonValue;
+    private bool _reasonInputActive => FTQReason != null ? FTQReason.Contains("FTC") : false;
+    private MudTextField<string?> reasonInput = default!;
+    private MudSelect<string> multiSelect = default!;
```
achobanov: Are these really necessary?

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Components/Actions/EliminationForms/FailedToQualifyForm.razor
```diff
@@ -0,0 +1,102 @@
...
+            if (i == 0 || i == codesCount - 1)
+            {
+                parsedCodes += code.ToString();
+            }
+            else
+            {
+                parsedCodes += code.ToString() + "+";
+            }
+        }
+        return parsedCodes;
```
achobanov: Can be replaced by a single `string.Join` call.

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Components/Actions/EliminationForms/FailedToQualifyForm.razor
```diff
@@ -0,0 +1,102 @@
...
+        }
+        else
+        {
+            await _participationBehind.RevokeQualification(tandemNumber, QualificationRevokeType.FailToQualify, null, codes.ToArray());
+        }
+    }
+
+    public async Task Restore()
+    {
+        Participation? participation = _participationBehind.SelectedParticipation;
```
achobanov: In what way is this variable helpful?

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Components/Actions/EliminationForms/FinishNotRankedForm.razor
```diff
@@ -0,0 +1,50 @@
+﻿@using NTS.Domain.Core.Aggregates.Participations
+@using NTS.Judge.Blazor.Enums
+@using NTS.Judge.Blazor.Ports
+@using System.Diagnostics
+
+@inherits NotComponent
+
+
```
achobanov, line 9: empty lines

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Components/Actions/EliminationForms/FinishNotRankedForm.razor
```diff
@@ -0,0 +1,50 @@
...
+    @Localizer.Get("Restore")
+</MudButton>
+
+@code {
+    [Inject]
+    IParticipationBehind _participationBehind { get; set; } = default!;
+    [Parameter]
+    public FinishedNotRanked? FinishedNotRanked { get; set; } = default!;
+    [Parameter]
+    public EventCallback OnSubmission { get; set; }
```
achobanov: Isn't `OnSubmission` a bit generic? Furthermore it's not entirely accurate, because you're only invoking it in one of the submission scenarios - Revoke.

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Components/Actions/EliminationForms/FinishNotRankedForm.razor
```diff
@@ -0,0 +1,50 @@
...
+        await _participationBehind.RevokeQualification(tandemNumber, QualificationRevokeType.FinishNotRanked, reason);   
+    }
+
+    public async Task Restore()
+    {
+        Participation? participation = _participationBehind.SelectedParticipation;
+        var tandemNumber = participation!.Tandem.Number;
+        await OnSubmission.InvokeAsync();
+        await _participationBehind.RestoreQualification(tandemNumber);
+    }
```
achobanov: `OnSubmission` and `Restore` seem to be completely identical. What can we do about that?

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Components/Actions/EliminationForms/FinishNotRankedForm.razor
```diff
@@ -0,0 +1,50 @@
...
+
+    private string? Reason
+    {
+        get => FinishedNotRanked != null ? FinishedNotRanked.Complement : _value;
+        set { _value = value; }
+    }
+    private string? _value;
+    private MudTextField<string?> reasonInput = default!;
+
+    public async Task Revoke()
```
achobanov: Should probably be renamed to `Eliminate`

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Components/Actions/EliminationsPanel.razor
```diff
@@ -1,133 +1,68 @@
...
 @code {
     [Inject]
     IParticipationBehind _participationBehind { get; set; } = default!;
-
+    NotQualified? _notQualified => _participationBehind.SelectedParticipation?.NotQualified;
+    private string? _buttonValue {
+        get => _value != null ? _value : _notQualified?.EliminationCode;
+        set { _value = value; }
+    }
+    private string? _value;
```
achobanov: Naming is confusing here as well. Think about how we can improve it:
- What does `_value` contain?
- What does `_buttonValue` contain?

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Components/Actions/EliminationsPanel.razor
```diff
@@ -1,133 +1,68 @@
...
     private MudToggleGroup<string> toggleGroup = default!;
     private Participation _selectedParticipation = default!;
```
achobanov: Are these necessary?

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Components/Actions/EliminationsPanel.razor
```diff
@@ -1,133 +1,68 @@
...
-        }
-        if (_qualificationRevokeType == null)
-        {
-            await _participationBehind.RestoreQualification(data.TandemNumber);
-        }
-        else
-        {
-            await _participationBehind.RevokeQualification(data.TandemNumber, (QualificationRevokeType)_qualificationRevokeType, data.Reason, data.Codes.ToArray());
-        }
+        _buttonValue = value;
```
achobanov: Can you not achieve this by simply using `@bind-Value=_buttonValue`?

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Components/Actions/ManualTimeSnapshot.razor
```diff
@@ -49,6 +56,7 @@
         if (TimeValue != "00:00:00")
```
achobanov: Line 34 comment: is it tested? If so, remove it.

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Components/Actions/ManualTimeSnapshot.razor
```diff
@@ -49,6 +56,7 @@
...
         {
             var convertedInput = TimeSpan.Parse(TimeValue);
+            TimeValue = "00:00:00";
```
achobanov: magic strings

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Components/Actions/ManualTimeSnapshot.razor
```diff
@@ -49,6 +56,7 @@
...
             var time = DateTime.Today + convertedInput;
             var snapshot = CreateSnapshot(time);
             _participationBehind?.Process(snapshot);
```
achobanov: `_participationBehind` should not be null here.

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Components/Actions/ManualTimeSnapshot.razor
```diff
@@ -17,6 +19,11 @@
     private MudTextField<string> timeInput = default!;
     private string TimeValue = "00:00:00";
```
achobanov: magic

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Components/Actions/ManualTimeSnapshot.razor
achobanov: also magic

src-v2/Judge/NTS.Judge/Adapters/Behinds/ParticipationBehind.cs
```diff
@@ -115,6 +120,8 @@ public async Task RevokeQualification(int number, QualificationRevokeType type,
             participation.FailToCompleteLoop(reason);
         }
         await _participationRepository.Update(participation);
+
+        EmitChange();
     }
```
achobanov, line 104: This is one big ass elimination method. Why don't we get rid of the last if-chain and split this up into named methods following the domain model?

src-v2/NTS.Domain.Core/Aggregates/Participations/NotQualified.cs
```diff
@@ -66,8 +71,10 @@ public FailedToQualify(params FTQCodes[] codes)
             throw new DomainException($"'Failed to Complete' requires a writen explanation from officials. Please provide 'complement'");
         }
         Codes = codes;
+        EliminationCode = FAILED_TO_QUALIFY;
```
achobanov: why don't you pass it to the base ctor instead?

src-v2/Not/Not/Extensions/EnumHelper.cs
```diff
@@ -7,8 +7,9 @@ public static class EnumHelper
     public static string GetDescription(this Enum value)
     {
         var type = value.GetType();
-        var descriptionAttribute = ReflectionHelper.GetEnumField(type, value)?.GetCustomAttributes(typeof(DescriptionAttribute), false)
-                                                             .FirstOrDefault() as DescriptionAttribute;
+        var descriptionAttribute = ReflectionHelper.GetEnumField(type, value)?
+                                                   .GetCustomAttributes(typeof(DescriptionAttribute), false)
+                                                   .FirstOrDefault() as DescriptionAttribute;
```
achobanov: Should be:
```
 var descriptionAttribute = ReflectionHelper
    .GetEnumField(type, value)
    ?.GetCustomAttributes(typeof(DescriptionAttribute), false)
    .FirstOrDefault() as DescriptionAttribute;
```

src-v2/Not/Not/Structures/SelectListModel.cs
```diff
@@ -4,6 +4,12 @@ namespace Not.Structures;
 
 public class SelectListModel
 {
+    public static IEnumerable<SelectListModel> SelectList<T>()
+    {
+        var enumValues = Enum.GetValues(typeof(T)).Cast<Enum>();
+        var selectItems = enumValues.Select(s => new SelectListModel(s.ToString(), s.GetDescription()));
+        return selectItems;
+    }
```
achobanov: What if `T` is not an enum. While constraining a generic param to enum type isn't entirely possible on compile time, there are 3 major things that we can do to prevent errors:
1. Constraint T to struct:
```
public static IEnumerable<SelectListModel> SelectList<T>()
    where T : struct
```
This prevents the method from compiling successfully if it's attempted to be used with reference types, aka classes. 

2. Verify that the type is enum - using `Type.IsEnum` property you can determine whether or not T is an Enum type and throw meaningful error if it is not. This is an OK solution, since typeof(T) isn't an expensive reflection operation. 

3. Most notably - you can indicate that T is enum in the method name.

achobanov (changes requested):

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Components/Actions/EliminationForms/FailedToQualifyForm.razor
StoyanGoranov: The reference of the *multiSelect* is used to get an IEnumerable of selected values to parse them to Enums.
The *reasonInput* ref I've removed.

src-v2/NTS.Domain.Core/Aggregates/Participations/NotQualified.cs
StoyanGoranov: Because for correct deserialization the ctor needs to be decorated with the [JsonConstructor] attribute. However, it is already used once in this class. Another usage creates ambiguity and results in runtime errors.

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Components/Actions/EliminationForms/FailedToQualifyForm.razor
StoyanGoranov: Actually, it isn't necessary after all.

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Components/Actions/EliminationsPanel.razor
StoyanGoranov: When this is done, a bug appears that causes the application to freeze when Eliminate button is clicked. When reverting to separate Value and ValueChanged the bug doesn't occur.

src-v2/NTS.Domain.Core/Aggregates/Participations/NotQualified.cs
achobanov: "results in runtime errors" I presume you meant "compile-time errors"? 

Using another `JsonConstructor` attribute is a nonsense, because that attribute is used to select a ctor that used during deserialization to instantiate the object. If there are two attributes how is the deserializer to know which one to select? 

That said this has nothing to do with base ctor. If this ctor isn't used during deserialization, then no changes to it will affect deserialization. For example if you set `EliminationCode`  to `Retired`, but the the serialized data does not have `Retired` value, your object after deserialization will contain whatever the value is in the serialized data. It's not going to be `Retired`, because that's not the ctor that the deserializer used. All of this is a long way of saying, you seem to be confusing constructors.

In fact, as is now, `NotQualified` has empty protected ctor. This then allows children to be instantiated without `EliminationCode` which could be a problem.

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Components/Actions/EliminationForms/DisqualifyForm.razor
```diff
@@ -20,29 +20,26 @@
...
     public Disqualified? Disqualified { get; set; }
     [Parameter]
-    public EventCallback OnSubmission { get; set; }
+    public EventCallback OnRestore { get; set; }
+    [Parameter]
+    public EventCallback OnEliminate { get; set; }
+    private string? Reason { get; set; }
 
-    private string? Reason 
+    protected override void OnParametersSet()
```
achobanov, line 22: This is a very good improvement! Good job

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Components/Actions/EliminationForms/DisqualifyForm.razor
```diff
@@ -20,29 +20,26 @@
...
-    private MudTextField<string?> reasonInput = default!;
 
-    public async Task Revoke()
+    public async Task Eliminate()
     {
         Participation? participation = _participationBehind.SelectedParticipation;
         var tandemNumber = participation!.Tandem.Number;
-        var reason = Reason;
-        await _participationBehind.RevokeQualification(tandemNumber, QualificationRevokeType.Disqualify, reason);
+        await _participationBehind.Disqualify(tandemNumber, Reason);
```
achobanov: Do you need to provide the tandemNumber as parameter here?

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Components/Actions/EliminationForms/FailedToQualifyForm.razor
```diff
@@ -34,69 +33,57 @@
     [Parameter]
     public FailedToQualify? FailedToQualify { get; set; }
     [Parameter]
-    public EventCallback OnSubmission { get; set; }
+    public EventCallback OnRestore { get; set; }
+    [Parameter]
+    public EventCallback OnEliminate { get; set; }
 
-    private string? FTQReason
+    private IEnumerable<string>? FTQReason { get; set; }
```
achobanov: Does it have to be a property? Also `codes` seems like a better name to me. You're already in FTQ form, no need to provide extra context, unless necessary.

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Components/Actions/EliminationForms/FailedToQualifyForm.razor
```diff
@@ -34,69 +33,57 @@
...
+    public string? Reason { get; set; }
+    private bool _reasonInputState;
```
achobanov: Is this necessary? Why not simple render reason input if there is `Complement` value?

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Components/Actions/EliminationForms/FailedToQualifyForm.razor
```diff
@@ -34,69 +33,57 @@
...
 
-    public async Task Revoke()
+    public async Task Eliminate()
     {
-        Participation? participation = _participationBehind.SelectedParticipation;
-        var tandemNumber = participation!.Tandem.Number;
-        var selectedValues = multiSelect.SelectedValues;
-        var reason = reasonInput.Value;
+        var tandemNumber = _participationBehind!.SelectedParticipation!.Tandem.Number;
         List<FTQCodes> codes = new List<FTQCodes>();
```
achobanov: var

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Components/Actions/EliminationForms/FailedToQualifyForm.razor
```diff
@@ -6,21 +6,20 @@
 
 @inherits NotComponent
 
-<MudSelect @ref="multiSelect" T="string" Label="Failed to Qualify Reason" MultiSelection="true" Delimiter="+" Value="@FTQReason" Placeholder="Select a reason">
-    @foreach (var failToQualifyCode in SelectListModel.SelectList<FTQCodes>())
+<MudSelect T="string" Label="Failed to Qualify Reason" MultiSelection="true" Delimiter="+" SelectedValues="@FTQReason" SelectedValuesChanged="SetInputState" Placeholder="Select a reason">
```
achobanov: Live is way too long

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Components/Actions/EliminationForms/FailedToQualifyForm.razor
achobanov: Furthermore why not use the enum directly, rather than it's string values?

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Components/Actions/EliminationForms/WithdrawForm.razor
```diff
@@ -19,20 +19,20 @@
...
+    public async Task Eliminate()
     {
         Participation? participation = _participationBehind.SelectedParticipation;
         var tandemNumber = participation!.Tandem.Number;
-        await _participationBehind.RevokeQualification(tandemNumber, QualificationRevokeType.Withdraw);
+        await _participationBehind.Withdraw(tandemNumber);
+        await OnEliminate.InvokeAsync();
     }
 
     public async Task Restore()
```
achobanov: It's getting much better, but those hooks and methods have identical signature and usage across all elimination forms. What can we do about (except merging the forms, obviously)?

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Components/Actions/EliminationsPanel.razor
```diff
@@ -9,46 +9,48 @@
...
     [Inject]
     IParticipationBehind _participationBehind { get; set; } = default!;
     NotQualified? _notQualified => _participationBehind.SelectedParticipation?.NotQualified;
-    private string? _buttonValue {
-        get => _value != null ? _value : _notQualified?.EliminationCode;
-        set { _value = value; }
+    private string? _toggledButtonValue {
+        get => _eliminationCodeValue != null ? _eliminationCodeValue : _notQualified?.EliminationCode;
+        set { _eliminationCodeValue = value; }
     }
```
achobanov: Looks cleaner:
```
private string? _toggledButtonValue
{
    get => _eliminationCodeValue != null ? _eliminationCodeValue : _notQualified?.EliminationCode;
    set => _eliminationCodeValue = value;
}
```

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Components/Actions/EliminationsPanel.razor
```diff
@@ -9,46 +9,48 @@
...
-    private string? _value;
-    private MudToggleGroup<string> toggleGroup = default!;
-    private Participation _selectedParticipation = default!;
+    private string? _eliminationCodeValue;
```
achobanov: Naming isn't still there. While technically speaking correct, it does not convey its actual purpose.  Think about why this property exists separately from `_toggleButtonValue`

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Components/Actions/EliminationsPanel.razor
achobanov: I don't think that's what was causing the bug. 

An interesting fact is that if you try to use `@bind-Value` while having `ValueChanged` set as well you get a compiler error that `ValueChanged` was used more than once. That is because `@bind-Value` is essentially a short-hand syntax for what you have made yourself:
1. Sets `Value` to the property
2. Creates an EventCallback that updates the value of the property
3. Sets `ValueChanged` to the previously created event callback. 

That is why you need to use a property with public getter and setter in order to use `@bind-Value`.

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Components/Actions/EliminationsPanel.razor
```diff
@@ -58,11 +60,18 @@
 
     public void OnToggle(string value)
     {
-        _buttonValue = value;
+        _toggledButtonValue = value;
+    }
+
+    public async Task RestoreAndUntoggle()
```
achobanov: Having `and` in the method/object names isn't ideal. Isn't there a way for the elimination forms to keep ownership of the restore feature and still be reused?

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Components/Actions/EliminationsPanel.razor
```diff
@@ -58,11 +60,18 @@
...
+    {
+        var tandemNumber = _participationBehind!.SelectedParticipation!.Tandem.Number;
+        await _participationBehind.RestoreQualification(tandemNumber);
+        _toggledButtonValue = null;
     }
 
-    public void Untoggle()
+    public void ResetButtonPrivateValue()
```
achobanov: Name is confusing - `OnEliminate` would be better here as well, don you think?

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Components/Actions/ManualTimeSnapshot.razor
```diff
@@ -16,8 +16,9 @@
 @code {
     [Inject]
     IParticipationBehind _participationBehind { get; set; } = default!;
-    private MudTextField<string> timeInput = default!;
-    private string TimeValue = "00:00:00";
+    private const string NULL_TIME = "00:00:00";
```
achobanov: DEFAULT_VALUE would be better name imo

src-v2/NTS.Domain.Core/Aggregates/Participations/NotQualified.cs
```diff
@@ -73,6 +73,12 @@ public FailedToQualify(params FTQCodes[] codes)
         Codes = codes;
         EliminationCode = FAILED_TO_QUALIFY;
     }
+
+    public FailedToQualify(string complement ,params FTQCodes[] codes) : base(complement, FAILED_TO_QUALIFY)
+    {
+        Codes = codes;
+        EliminationCode = FAILED_TO_QUALIFY;
```
achobanov: Is this line necessary? Think about how you set the `EliminationCode` here. Related to [previous comment](https://github.com/Not-Endurance/not-timing-system/pull/270#discussion_r1771275668)

src-v2/NTS.Domain.Core/Aggregates/Participations/Participation.cs
```diff
@@ -107,9 +107,9 @@ public void FailToQualify(params FTQCodes[] codes)
         RevokeQualification(new FailedToQualify(codes));
     }
 
-    public void FailToCompleteLoop(string reason)
+    public void FailToCompleteLoop(string reason, params FTQCodes[] codes)
     {
-        RevokeQualification(new FailedToQualify(reason));
+        RevokeQualification(new FailedToQualify(reason, codes));
     }
```
achobanov: I'm thinking that there isn't too much benefit to have a separate method for `FailToCompleteLoop` since it's inherently a `FailToQualify`. What do you thknk?

src-v2/Not/Not/Structures/SelectListModel.cs
```diff
@@ -4,7 +4,8 @@ namespace Not.Structures;
 
 public class SelectListModel
 {
-    public static IEnumerable<SelectListModel> SelectList<T>()
+    public static IEnumerable<SelectListModel> SelectListFromEnum<T>()
```
achobanov: Isn't `SelectListModel.SelectListFromEnum` a bit redundant?

achobanov (changes requested):

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Components/Actions/EliminationsPanel.razor
```diff
@@ -0,0 +1,77 @@
...
+        </MudCollapse>
+        <MudCollapse Expanded="@(_toggledButtonValue == NotQualified.FINISHED_NOT_RANKED)">
+            <FinishNotRankedForm FinishedNotRanked="_notQualified as FinishedNotRanked"
+                                 OnRestore="RestoreAndUntoggle" OnEliminate="ResetButtonPrivateValue" />
+        </MudCollapse>
+        <MudCollapse Expanded="@(_toggledButtonValue == NotQualified.DISQUALIFIED)">
+            <DisqualifyForm Disqualified="_notQualified as Disqualified"
+                            OnRestore="RestoreAndUntoggle" OnEliminate="ResetButtonPrivateValue" />
+        </MudCollapse>
+        <MudCollapse Expanded="@(_toggledButtonValue.Contains(NotQualified.FAILED_TO_QUALIFY))">
```
achobanov: Why is this check only using `Contains`?

achobanov (changes requested):

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Components/Actions/EliminationsPanel.razor
achobanov: Curly bracket should be on line 50

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Components/Actions/EliminationsPanel.razor
achobanov: Still relevant. What's the reason for this field to exist and how can the name be improved to relay that reason?

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Components/Actions/EliminationForms/DisqualifyForm.razor
```diff
@@ -14,32 +15,18 @@
...
         Reason = Disqualified?.Complement;
     }
 
-    public async Task Eliminate()
+    public override async Task Eliminate()
     {
-        Participation? participation = _participationBehind.SelectedParticipation;
-        var tandemNumber = participation!.Tandem.Number;
-        await _participationBehind.Disqualify(tandemNumber, Reason);
+        await _participationBehind.Disqualify(Reason);
```
achobanov, line 30: What if reason is null here?

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Components/Actions/EliminationForms/FailedToQualifyForm.razor
```diff
@@ -4,18 +4,19 @@
 @using NTS.Judge.Blazor.Ports
 @using Not.Structures
 
-@inherits NotComponent
+@inherits EliminationForm
 
-<MudSelect T="string" Label="Failed to Qualify Reason" MultiSelection="true" Delimiter="+" SelectedValues="@FTQReason" SelectedValuesChanged="SetInputState" Placeholder="Select a reason">
-    @foreach (var failToQualifyCode in SelectListModel.SelectListFromEnum<FTQCodes>())
+<MudSelect T="FTQCodes" Label="Failed to Qualify Reason" MultiSelection="true" Delimiter="+"
+    SelectedValues="@Codes" SelectedValuesChanged="SetInputState" Placeholder="Select a reason">
```
achobanov, line 10: `T="FTQCodes"` isn't  necessary as it can be auto-inferred by the usage of `SelectedValues`.

Furthermore you can use `@bind-SelectedValues` and remove `SelectedValuesChanged` here as well.

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Components/Actions/EliminationForms/FailedToQualifyForm.razor
```diff
@@ -4,18 +4,19 @@
...
+    @foreach (var failToQualifyCode in SelectListModel.FromEnum<FTQCodes>())
     {
         <MudSelectItem T="string" Value="@failToQualifyCode.Value">
```
achobanov: Honestly I'm surprised this is working now, because I would expect that MudSelect and MudSelectItem should have the same type of T. Anyhow, doesn't it make more sense to have the value of SelectListModel to be int. Enum types are int by default, so they would be interchangeable.

Furthermore you should be able to remove T property here as well

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Components/Actions/EliminationForms/FailedToQualifyForm.razor
```diff
@@ -4,18 +4,19 @@
...
             @failToQualifyCode.Value
```
achobanov: Use Mud's typography instead of directly rendering text.

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Components/Actions/EliminationForms/FailedToQualifyForm.razor
```diff
@@ -4,18 +4,19 @@
...
             <div class="font-size-10">@failToQualifyCode.Description</div>
```
achobanov: Use Mud typography instead of custom font sizes.

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Components/Actions/EliminationForms/FailedToQualifyForm.razor
```diff
@@ -28,62 +29,41 @@
...
-    IParticipationBehind _participationBehind { get; set; } = default!;
     [Parameter]
     public FailedToQualify? FailedToQualify { get; set; }
-    [Parameter]
-    public EventCallback OnRestore { get; set; }
-    [Parameter]
-    public EventCallback OnEliminate { get; set; }
 
-    private IEnumerable<string>? FTQReason { get; set; }
+    private IEnumerable<FTQCodes> Codes { get; set; } = new List<FTQCodes>();
```
achobanov: As this is a collection that is modified by this component it can be defined as a List directly

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Components/Actions/EliminationForms/FailedToQualifyForm.razor
```diff
@@ -28,62 +29,41 @@
...
     {
-        var tandemNumber = _participationBehind!.SelectedParticipation!.Tandem.Number;
-        List<FTQCodes> codes = new List<FTQCodes>();
-        foreach (var value in FTQReason)
+        var codes = new List<FTQCodes>();
+        foreach (var value in Codes)
         {
-            codes.Add((FTQCodes)Enum.Parse(typeof(FTQCodes), value));
+            codes.Add(value);
         }
```
achobanov: Is this necessary?

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Components/Actions/EliminationForms/FailedToQualifyForm.razor
```diff
@@ -28,62 +29,41 @@
...
         if (codes.Contains(FTQCodes.FTC))
         {
-            await _participationBehind.FailToCompleteLoop(tandemNumber, Reason, codes.ToArray());
+            await _participationBehind.FailToQualify(Reason, codes.ToArray());
         }
         else
         {
-            await _participationBehind.FailToQualify(tandemNumber, codes.ToArray());
+            await _participationBehind.FailToQualify(codes.ToArray());
         }
```
achobanov: We discussed that we don't need two separate methods for FTQ in Participation, right?

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Components/Actions/EliminationsPanel.razor
```diff
@@ -47,31 +47,30 @@
...
+        _toggledButtonValue = null;
     }
 
-    public async Task RestoreAndUntoggle()
+    public async Task RestoreHandler()
     {
-        var tandemNumber = _participationBehind!.SelectedParticipation!.Tandem.Number;
-        await _participationBehind.RestoreQualification(tandemNumber);
-        _toggledButtonValue = null;
+        await _participationBehind.RestoreQualification();
```
achobanov: Isn't there a better place for this?

src-v2/Judge/NTS.Judge/Adapters/Behinds/ParticipationBehind.cs
```diff
@@ -87,49 +87,54 @@ public async Task Update(IPhaseState state)
         await _participationRepository.Update(participation);
     }
 
-    public async Task Withdraw(int number)
+    public async Task Withdraw()
     {
-        var participation = Participations.FirstOrDefault(x => x.Tandem.Number == number);
+        var tandemNumber = SelectedParticipation!.Tandem.Number;
+        var participation = Participations.FirstOrDefault(x => x.Tandem.Number == tandemNumber);
```
achobanov: Is this necessary? Applies to all elimination methods

achobanov (changes requested):
"Address review #" isn't a very good commit message as it does nothing to explain what changes are occurring in the code. You can group thinks like variable/field renaming and formatting, but other changes should be separate:
- Extract base EliminationForm
- Use FTQCodes enum directly in FailToQualifyForm
- Remove tandem number argument from elimination methods in behind
etc

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Components/Actions/EliminationForms/FailedToQualifyForm.razor
achobanov: "Honestly I'm surprised this is working now" - it's no longer working when I update to dotnet 8. It's causing the application to hang on a white screen with no errors whatsoever. You can see how I've fixed it [here](https://github.com/Not-Endurance/not-timing-system/commit/c9c52e5a5cd74b2f7924ef34927ccf313d2bee62)

achobanov: Don't fix it in this PR.

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Components/Actions/EliminationForms/FailedToQualifyForm.razor
StoyanGoranov: Removed the calls from FailedToQualify.razor and ParticipationBehind.
`public void FailToQualify(params FTQCodes[] codes)` at Participation.cs is
still in use at ParticipationFactory.cs

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Components/Actions/EliminationForms/DisqualifyForm.razor
StoyanGoranov: This should be validated. Adding To Dos for validation. An abstraction of some sort for NotForm<T> is probably needed to enable the use of RegisterValidationInjectors for the EliminationForms as well.

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Components/Actions/EliminationForms/FailedToQualifyForm.razor
achobanov: What about this one?

achobanov (approved):

PR #280, opened 2024-08-23
nts/startlist

### LINKED ISSUE #252: Create Startlist panel
## Overview
Component that renders and is able to print the current Startlist.

StoyanGoranov:
Closes: https://github.com/Not-Endurance/not-timing-system/issues/252


src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/StartlistPage.razor
```diff
@@ -0,0 +1,83 @@
...
+@using System.Collections
+@using Not.Blazor.TM
+@using Not.Extensions
+
+@attribute [Route(STARTLIST_PAGE)]
+@inherits NotComponent
+
+<NotH1>@Localizer.Get("Startlist")</NotH1>
+
+<MudGrid Style="margin-top:0 !important;">
```
achobanov: Grid isn't necessary here

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/StartlistPage.razor
```diff
@@ -0,0 +1,83 @@
...
+    <MudTabs @ref="@Tabs" PanelClass="px-4 py-6" Elevation="4"
+        Rounded ApplyEffectsToContainer>
```
achobanov: Should be on upper line

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/StartlistPage.razor
```diff
@@ -0,0 +1,83 @@
...
+                        </thead>
+                        <tbody>
+                            @foreach (var row in groupedData)
+                            {
+                                <tr>
+                                    @foreach (var x in row.ToString().Split(","))
+                                    {
+                                        <td>@x</td>
+                                    }
+                                    <td><NotTimer StartTime="@row.StartIn()" StopTimerAtTime="-15"/></td>
```
achobanov: Lets avoid nesting elements on the same line.

src-v2/Judge/NTS.Judge.Blazor/Ports/IStartlistBehind.cs
```diff
@@ -0,0 +1,12 @@
+﻿using Not.Blazor.Ports.Behinds;
+using Not.Injection;
+using NTS.Domain.Core.Objects;
+
+namespace NTS.Judge.Blazor.Ports;
+
+public interface IStartlistBehind : IObservableBehind, ISingletonService
+{
+    StartList? StartList { get; }
+    List<Start> UpcomingStarts { get; }
```
achobanov: Simply Upcoming and History should be enough

src-v2/Judge/NTS.Judge/Adapters/Behinds/StartlistBehind.cs
```diff
@@ -0,0 +1,53 @@
...
+    public TimerUtility _timer { get; set; }
+
+    protected override async Task<bool> PerformInitialization(params IEnumerable<object> arguments)
+    {
+        var participations = await _participationRepository.ReadAll();
+        var startlist = new StartList(participations);
+        GuardHelper.ThrowIfDefault(startlist);
+        UpcomingStarts = startlist.UpcomingStarts;
+        StartHistory = startlist.PreviousStarts;
+        StartList = startlist;
```
achobanov: What is the point of exposing 3 properties that overlap?

src-v2/NTS.Domain.Core/Objects/Start.cs
```diff
@@ -0,0 +1,75 @@
...
+                    {
+                        var upcomingStart = new Start(participation.Combination.Name, participation.Combination.Number, index + 1, phases[index].Length, participation.Phases.Distance, phase.StartTime!);
+                        UpcomingStarts.Add(upcomingStart);
+                    }
+                }
+            }
+        }
+    }
+    public List<Start> UpcomingStarts { get; set; } = new List<Start>();
+    public List<Start> PreviousStarts { get; set; } = new List<Start>();
```
achobanov: Domain models should not expose mutable collections unless it's specifically intended for outsiders to modify them

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Startlist/StartlistHistory.razor
```diff
@@ -0,0 +1,102 @@
...
+        @foreach (var tab in UserTabs)
+        {
+            <MudTabPanel ID="@tab.Id" Text="@tab.Label">
+                <ChildContent>
+                    <MudDataGrid @ref="dataGrid" MultiSelection="true" Items="@Items.Where(g=>g.TotalDistance == tab.Key)" Dense="true"
+                                 Hideable="true" Groupable="true" GroupExpanded="true">
+                        <ToolBarContent>
+                            <MudDivider DividerType="DividerType.FullWidth" Absolute></MudDivider>
+                            <th class="group-header">Participant</th>
+                            <MudDivider DividerType="DividerType.FullWidth" Absolute></MudDivider>
```
achobanov: Don't think you need two dividers in order to center something,

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Startlist/StartlistHistory.razor
```diff
@@ -0,0 +1,102 @@
...
+}
+else
+{
+    <MudTabs @ref="@Tabs" PanelClass="px-4 py-6" Elevation="4" Rounded ApplyEffectsToContainer>
+        @foreach (var tab in UserTabs)
+        {
+            <MudTabPanel ID="@tab.Id" Text="@tab.Label">
+                <ChildContent>
+                    <MudDataGrid @ref="dataGrid" MultiSelection="true" Items="@Items.Where(g=>g.TotalDistance == tab.Key)" Dense="true"
+                                 Hideable="true" Groupable="true" GroupExpanded="true">
```
achobanov: What's the Data grid advantage over ToString as we've used before?

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Startlist/StartlistHistory.razor
```diff
@@ -0,0 +1,102 @@
...
+    </MudTabs>
+
+}
+
+@code{
+    [Parameter]
+    public IEnumerable<Start> Items { get; set; } = new List<Start>();
+
+    MudDataGrid<Start> dataGrid = default!;
+    MudTabs Tabs = default!;
```
achobanov: What's the purpose of this property?

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Startlist/StartlistHistory.razor
```diff
@@ -0,0 +1,102 @@
...
+    <MudText Typo="Typo.body1">Start history is empty.</MudText>
+}
+else
+{
+    <MudTabs @ref="@Tabs" PanelClass="px-4 py-6" Elevation="4" Rounded ApplyEffectsToContainer>
+        @foreach (var tab in UserTabs)
+        {
+            <MudTabPanel ID="@tab.Id" Text="@tab.Label">
+                <ChildContent>
+                    <MudDataGrid @ref="dataGrid" MultiSelection="true" Items="@Items.Where(g=>g.TotalDistance == tab.Key)" Dense="true"
```
achobanov: Are we grouping for aesthetics, convenience or by business requirement?

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Startlist/StartlistHistory.razor
```diff
@@ -0,0 +1,102 @@
...
+        padding-left:16px;
+        background-color:rgb(245, 245, 245);
+    }
+    .mud-toolbar{
+        height:38.8px !important;
+    }
+    .mud-table-toolbar{
+        padding-left: 0px !important;
+        padding-right: 0px !important;
+    }
```
achobanov: Why do we need custom styles? Going forward, I want any custom styles to have specific comments describing their purpose.

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Startlist/StartlistPage.razor
```diff
@@ -0,0 +1,32 @@
...
+
+
+<MudTabs Elevation="2" Rounded="true" ApplyEffectsToContainer="true" PanelClass="pa-6">
+    <MudTabPanel Text="Upcoming Starts">
+        <UpcomingStarts Items="UpcomingStartsList"></UpcomingStarts>
+    </MudTabPanel>
+    <MudTabPanel Text="Start History">
+        <StartlistHistory Items="StartHistoryList"></StartlistHistory>
+    </MudTabPanel>
+</MudTabs>
```
achobanov: Tabs within tabs doesn't look good. Use `NotPage` to conform to Dashboard pages.

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Startlist/TabView.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace NTS.Judge.Blazor.Pages.Dashboard.Startlist;
+public class TabView
```
achobanov: Don't like the naming, we're not using Views atm. 

Furthermore is this the best place for a generic component model?

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Startlist/UpcomingStarts.razor
```diff
@@ -0,0 +1,79 @@
...
+{
+    <MudTabs @ref="@Tabs" PanelClass="px-4 py-6" Elevation="4" Rounded ApplyEffectsToContainer>
+        @foreach (var tab in UserTabs)
+        {
+            <MudTabPanel ID="@tab.Id" Text="@tab.Label">
+                <ChildContent>
+                    <MudSimpleTable Dense="@dense" Hover="@hover" Bordered="@bordered" Striped="@striped" Style="overflow-x: auto;">
+                        <thead>
+                            <tr>
+                                @foreach (var h in headings)
```
achobanov: Is this a good name?

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Startlist/UpcomingStarts.razor
```diff
@@ -0,0 +1,79 @@
...
+            </MudTabPanel>
+        }
+    </MudTabs>
+}
+  
+
+
+@code {
+    [Parameter]
+    public IEnumerable<Start> Items { get; set; } = new List<Start>();
```
achobanov: Is this a good name?

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Startlist/UpcomingStarts.razor
```diff
@@ -0,0 +1,79 @@
...
+                                    <th>@h</th>
+                                }
+                            </tr>
+                        </thead>
+                        <tbody>
+                            @foreach (var row in Items)
+                            {
+                                @if(tab.Label.Contains(row.TotalDistance.ToString())){
+                                    <tr>
+                                        @foreach (var x in row.ToString().Split(", "))
```
achobanov: Is this a good name?

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Startlist/UpcomingStarts.razor
```diff
@@ -0,0 +1,79 @@
...
+                        <thead>
+                            <tr>
+                                @foreach (var h in headings)
+                                {
+                                    <th>@h</th>
+                                }
+                            </tr>
+                        </thead>
+                        <tbody>
+                            @foreach (var row in Items)
```
achobanov: Is this a good name?

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Startlist/UpcomingStarts.razor
```diff
@@ -0,0 +1,79 @@
...
+    [Parameter]
+    public IEnumerable<Start> Items { get; set; } = new List<Start>();
+
+    MudTabs Tabs = default!;
+    List<TabView> UserTabs = new();
+    int UserIndex;
+    bool dense = false;
+    bool hover = false;
+    bool bordered = false;
+    bool striped = false;
```
achobanov: Are these necessary?

src-v2/Judge/NTS.Judge.Blazor/Ports/IStartlistBehind.cs
```diff
@@ -0,0 +1,12 @@
+﻿using Not.Blazor.Ports.Behinds;
+using Not.Injection;
+using NTS.Domain.Core.Objects;
+
+namespace NTS.Judge.Blazor.Ports;
+
+public interface IStartlistBehind : IObservableBehind, ISingletonService
```
achobanov: Can probably be split into smaller segregated interfaces.

src-v2/Judge/NTS.Judge/Adapters/Behinds/StartlistBehind.cs
```diff
@@ -0,0 +1,53 @@
...
+using NTS.Domain.Core.Entities;
+using NTS.Domain.Core.Objects;
+using NTS.Judge.Blazor.Ports;
+using System.Diagnostics;
+using System.Timers;
+
+namespace NTS.Judge.Adapters.Behinds;
+public class StartlistBehind : ObservableBehind, IStartlistBehind
+{
+    private const double TimerTickInMilliseconds = 60000;
```
achobanov: Isn't there a better object to describe a span of time?

Furthermore - refresh on every 60 seconds?

src-v2/NTS.Domain.Core/Objects/Start.cs
```diff
@@ -0,0 +1,75 @@
...
+                {
+                    var index = phases.IndexOf(phase);
+                    if (now - phase.StartTime > TimeSpan.FromMinutes(15))
+                    {
+                        var previousStart = new Start(participation.Combination.Name, participation.Combination.Number, index + 1, phases[index].Length, participation.Phases.Distance, phase.StartTime!);
+                        PreviousStarts.Add(previousStart);
+                    }
+                    else
+                    {
+                        var upcomingStart = new Start(participation.Combination.Name, participation.Combination.Number, index + 1, phases[index].Length, participation.Phases.Distance, phase.StartTime!);
```
achobanov: This instantiation is an exact copy of of the previous one. What are we doing with copy pase code?

achobanov: Furthermore it's very hard to read. It isn't at all obvious what are you using index for.

src-v2/NTS.Domain.Core/Objects/Start.cs
```diff
@@ -0,0 +1,75 @@
...
+                        UpcomingStarts.Add(upcomingStart);
+                    }
+                }
+            }
+        }
+    }
+    public List<Start> UpcomingStarts { get; set; } = new List<Start>();
+    public List<Start> PreviousStarts { get; set; } = new List<Start>();
+}
+public class Start
```
achobanov: Either separate in two files or move as an internal class to the `Startlist`, which you'll be able to use like `Startlist.Entry`.

src-v2/NTS.Domain.Core/Objects/Start.cs
```diff
@@ -0,0 +1,75 @@
...
+    {
+        if(StartAt == null) return string.Empty;
+        if (StartAt > Timestamp.Now())
+        {
+            return (StartAt - Timestamp.Now()).ToString();
+        }
+        else
+        {
+            return "-" + (StartAt - Timestamp.Now()).ToString();
+        }
```
achobanov: You can use just subtraction, same as 
> 5 - 7 = -2

src-v2/NTS.Domain.Core/Objects/Start.cs
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    public override string ToString()
+    {
+        var result = $"{Number}, {Athlete}, {"#".Localize()}{PhaseNumber}: {Distance}km, {StartAt}";
```
achobanov: We're using `DomainEntity.Combine` in ToString.

src-v2/Not/Not.Blazor/TM/NotTimer.razor
```diff
@@ -0,0 +1,63 @@
...
+    private MudTheme _theme = new MudTheme();
+    private TimeSpan _time = default!;
+    private string _displayTime = default!;
+    private string _color = default!;
+
+    protected override void OnParametersSet()
+    {
+        _time = TimeSpan.Parse(StartTime);
+        FormatTime();
+        _timer = new Timer(1000);
```
achobanov: Magic value.

achobanov: Also do you need to create a new timer every time a parameter is changed?

achobanov: Furthermore if we have a custom timer imlementation why not use that one for consistency?

src-v2/Not/Not.Blazor/TM/NotTimer.razor
```diff
@@ -0,0 +1,63 @@
...
+    private void OnTimerElapsed(object? sender, System.Timers.ElapsedEventArgs e)
+    {
+        _time = _time.Subtract(TimeSpan.FromSeconds(1));
+        FormatTime();
+        if (_time <= TimeSpan.FromMinutes(StopTimerAtTime))
+        {
+            _timer.Enabled = false;
+        }
+        _displayTime = _time.ToString().Replace("-", " - ");
+        InvokeAsync(StateHasChanged);
```
achobanov: Inherit `NotComponent` and use `Render` insitead of manually `InvokeAsync(StateHasChanged)`

src-v2/Not/Not/Extensions/DoubleExtension.cs
```diff
@@ -0,0 +1,14 @@
...
+{
+    public static double FloorWholeNumberToTens(this double number)
+    {
+        return (double)Math.Floor((decimal)number / 10) * 10;
+    }
+
+    public static double FloorWholeNumberToTens(this double? number)
+    {
+        if(number == null) throw new ArgumentNullException(nameof(number));
+        return (double)Math.Floor((decimal)number / 10) * 10;
```
achobanov: Isn't this a duplicate?

src-v2/Not/Not/Extensions/DoubleExtension.cs
```diff
@@ -0,0 +1,14 @@
+﻿namespace Not.Extensions;
+public static class DoubleExtension
+{
+    public static double FloorWholeNumberToTens(this double number)
+    {
+        return (double)Math.Floor((decimal)number / 10) * 10;
+    }
```
achobanov: This seems wrong on multiple accounts:
- Why would the method be named `FloorWholeNumber`? How do you floor an integer? An then accept double?
- Does it work correctly for values like `5.6`?

src-v2/Not/Not/TimerUtility/TimerUtility.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System.Timers;
+
+namespace Not.TimerUtility;
+public class TimerUtility
```
achobanov: We're avoiding the `Utility` name by convention. Utilities or Helpers should be very general purpose objects, in our case we use Helpers and they are mostly static.

As it currently is, I don't see much benefit from this class, as it barely does anything, whilst it's more restrictive than the System.Timers.Timer, which is why I presume you've opted to simply expose it. It is sometimes a good idea to restrict an API simply in order to enforce a consistent usage, but you're enforcing nothing when the underlying timer is exposed.

achobanov (changes requested):
I undstood your idea with the DataGrid and row grouping after I started the app, but that is not the desired effect. Seeing the starting times grouped by individual participation isn't of much value because we can already see that in the Dashboard. What could be useful is to compare the start times at a spicic stage for ALL participations (even across Competitions, regardless of the fact that they might have different Loops). Essentially you need to flip this up:
1. Group starts by Stage -lets define Stage as 1-based number of Phase for each individual Participation .So if we both participate, im doing 3 phases and you're only doing 2  and we are on our second Phase, we are both considered on our second Stage
2. The grouping logic is business logic, take it as que as to where it's supposed to be implemented.
3. I think you can simply use ToString to display the tabular data, same as with `CurdList`s instead of investing in DataGrid. A potential advantage of the grid implementation would be the ability to sort by column values, but I don't see that as very useful.

achobanov:
When you've made the changed you'll need to synch with develop before you merge for 2 reasons:
1. I've disabled merging when your branch is not synched, because sometimes even without conflicts you may have failures when you merge
2. I've added a Csharpier workflow, which you can see above, which runs automatically when commits are pushed to an open non-draft PR and formats the code. There was a bug in develop, which caused csharpier to fail, it has been fixed

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Startlist/StartlistHistory.razor
StoyanGoranov: DataGrid usage removed.

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Startlist/StartlistHistory.razor
StoyanGoranov: Changed the ordering by PhaseNumber

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Startlist/UpcomingStarts.razor
StoyanGoranov: Extracted MudSimpleTable usage in NotSimpleTable

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Startlist/StartlistHistory.razor
StoyanGoranov: This was used to style the DataGrid's ToolbarContent to make it look like a GroupHeader of the component

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Startlist/UpcomingStarts.razor
StoyanGoranov: Do you think the NotSimpleTable should inherit MudSimpleTable to avoid the need of these properties to be exposed explicitly as params?

src-v2/Not/Not/Extensions/DoubleExtension.cs
StoyanGoranov: The idea is to floor the Competition's TotalDistance i.e. 44.5 to 40 so it is displayed as a 40km bracket instead of the exact length.

src-v2/Judge/NTS.Judge/Adapters/Behinds/StartlistBehind.cs
StoyanGoranov: Unfortunately, TimeSpan.FromSeconds() cannot be assigned to a constant.

src-v2/Judge/NTS.Judge/Adapters/Behinds/StartlistBehind.cs
achobanov: I think this still stands. Further more properties in this class aren't encapsulated as they should be.

src-v2/NTS.Domain.Core/Objects/Start.cs
achobanov: Also still standing

src-v2/Judge/NTS.Judge.Blazor/Ports/IStartlistBehind.cs
achobanov: Still standing

src-v2/Judge/NTS.Judge/Adapters/Behinds/StartlistBehind.cs
achobanov: Why does it have to be a constant?

src-v2/NTS.Domain.Core/Objects/Start.cs
achobanov: Still standing

src-v2/Not/Not/Extensions/DoubleExtension.cs
achobanov: It isn't very obvious what this method is doing. Also I don't think this is a good candidate for a generic extension in Not. Not everything that's dealing with base types such as int, string etc should be in the Not project. A good question would be : how useful do you image this method to be in the average project? It seems to me that it's a piece of specific logic related our domain, so that's where I'd put it. Also keep in mind it can be used in other places than Start - for example the dashboard component and RanklistMenu are good candidates.

Regarding the method itself - both the naming and implementation are confusing. It shouldn't work with `double` or `decimal` as you are rounding away precision, not the other way around.

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Startlist/History.razor
```diff
@@ -0,0 +1,47 @@
...
+@using Not.Extensions
+
+@inherits NotComponent
+
+<NotTrueCenterText Typo="Typo.h5" Content="History" Class="mb-4" />
+
+<MudTabs @ref="@Tabs" PanelClass="px-4 py-6" Elevation="4" Rounded ApplyEffectsToContainer>
+    @foreach (var tab in UserTabs)
+    {
+        <MudTabPanel ID="@tab.Id" Text="@tab.Header">
```
achobanov: Not sure what's the point of this `TabModel`. Can't you simply use the values direclty? Id doesn't seem to be necessary.

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Startlist/Upcoming.razor
```diff
@@ -0,0 +1,46 @@
...
+@using Not.Extensions
+
+@inherits NotComponent
+
+<MudTabs @ref="@Tabs" PanelClass="px-4 py-6" Elevation="4" Rounded ApplyEffectsToContainer Class="@Class">
+    @foreach (var tab in UserTabs)
+    {
+        <MudTabPanel ID="@tab.Id" Text="@tab.Header">
+            <ChildContent>
+                <NotSimpleTable T="Start" Headings="@_tableHeadings" Items="Starts.Where(s => tab.Header.Contains(s.TotalDistance.ToString()))">
```
achobanov: This component seems practically identical to History... what do we do with identical code? On top of that, this isn't at all what we discussed as a feature.

src-v2/Judge/NTS.Judge/Adapters/Behinds/StartlistBehind.cs
```diff
@@ -1,53 +1,63 @@
...
         var participations = await _participationRepository.ReadAll();
-        var startlist = new StartList(participations);
-        GuardHelper.ThrowIfDefault(startlist);
-        UpcomingStarts = startlist.UpcomingStarts;
-        StartHistory = startlist.PreviousStarts;
-        StartList = startlist;
-        _timer = new TimerUtility(TimerTickInMilliseconds, upcomingStartExpire, () => true);
-        return UpcomingStarts.Any() && StartHistory.Any();
+        Startlist.AssignStarts(participations);
+        GuardHelper.ThrowIfDefault(Startlist);
```
achobanov: What's the point of this GuardHelper?

src-v2/Judge/NTS.Judge/Adapters/Behinds/StartlistBehind.cs
```diff
@@ -1,53 +1,63 @@
...
+        return Upcoming.Any() && History.Any();
     }
 
-    private void upcomingStartExpire(object? sender, ElapsedEventArgs e)
+    void upcomingStartExpire(object? sender, ElapsedEventArgs e)
```
achobanov: Methods should be PascalCase

src-v2/Judge/NTS.Judge/Adapters/Behinds/StartlistBehind.cs
```diff
@@ -1,53 +1,63 @@
...
 
 namespace NTS.Judge.Adapters.Behinds;
+
 public class StartlistBehind : ObservableBehind, IStartlistBehind
 {
-    private const double TimerTickInMilliseconds = 60000;
-    private const int ExpirationTimeOfPreviousStartsInMinutes = 15;
-    private readonly IRepository<Participation> _participationRepository;
+    const double TIMER_TICK_IN_MILLISECONDS = 1000;
+    const int EXPIRATION_TIME_OF_PREVIOUS_STARTS_IN_MINUTES = 15;
```
achobanov: The TimeSpan comment's point is to avoid having long and ugly constant names.

src-v2/Judge/NTS.Judge/Adapters/Behinds/StartlistBehind.cs
```diff
@@ -1,53 +1,63 @@
...
+        return Upcoming.Any() && History.Any();
     }
 
-    private void upcomingStartExpire(object? sender, ElapsedEventArgs e)
+    void upcomingStartExpire(object? sender, ElapsedEventArgs e)
     {
-        foreach (var start in UpcomingStarts.ToList())
+        foreach (var start in Upcoming.ToList())
         {
             var now = new Timestamp(DateTime.Now);
```
achobanov: You shouldn't need to instantiate a `Timestamp` here. You can make the check using existing timestamp operators or you can define operators for timestamp if necessary. See it's implementation to see how

src-v2/Judge/NTS.Judge/Adapters/Behinds/StartlistBehind.cs
```diff
@@ -1,53 +1,63 @@
...
-            if (now - start.StartAt > TimeSpan.FromMinutes(ExpirationTimeOfPreviousStartsInMinutes))
+            if (
+                now - start.StartAt
+                > TimeSpan.FromMinutes(EXPIRATION_TIME_OF_PREVIOUS_STARTS_IN_MINUTES)
+            )
             {
-                UpcomingStarts.Remove(start);
-                StartHistory.Add(start);
+                Startlist.Expire(start);
+                Startlist.OrderHistoryByAscending();
```
achobanov: THis is unnecessary.

src-v2/Judge/NTS.Judge/Adapters/Behinds/StartlistBehind.cs
```diff
@@ -1,53 +1,63 @@
...
             var now = new Timestamp(DateTime.Now);
-            if (now - start.StartAt > TimeSpan.FromMinutes(ExpirationTimeOfPreviousStartsInMinutes))
+            if (
+                now - start.StartAt
+                > TimeSpan.FromMinutes(EXPIRATION_TIME_OF_PREVIOUS_STARTS_IN_MINUTES)
+            )
             {
-                UpcomingStarts.Remove(start);
-                StartHistory.Add(start);
+                Startlist.Expire(start);
```
achobanov: I don't like this API. Either the `Startlist.cs` class should be aware of expiration and manage it's own time, OR it shouldn't care at all. Remember that at the core it should be the Domain Model that acts, rather than responds to other agents. The Behind is there just to integrate the Domain entity with infrastructure and UI.

src-v2/Judge/NTS.Judge/Factories/TimerFactory.cs
```diff
@@ -0,0 +1,19 @@
...
+    public static System.Timers.Timer CreateTimer(
+        double tickLength,
+        ElapsedEventHandler onTickHandler,
+        Func<bool> isActive
+    )
+    {
+        var _timer = new System.Timers.Timer(tickLength);
+        _timer.Elapsed += onTickHandler;
+        _timer.AutoReset = isActive();
+        _timer.Enabled = isActive();
```
achobanov: This is very confusing. Doens't the timer have a Start method which sets it's `Enabled` to true? What does `AutoReset` do? It's not at all clear how this is active method behaves for the consumer of this factory. Also why is `isActive` a Func?

src-v2/NTS.Domain.Core/Objects/Start.cs
```diff
@@ -56,20 +30,24 @@ public Start(Person athlete, int number, int loopNumber, double distance, double
...
         if (StartAt > Timestamp.Now())
         {
-            return (StartAt - Timestamp.Now()).ToString();
+            return (StartAt - Timestamp.Now())!.ToString();
         }
         else
         {
-            return "-" + (StartAt - Timestamp.Now()).ToString();
+            return "-" + (StartAt - Timestamp.Now())!.ToString();
         }
```
achobanov: Shouldn't have to use Timestamp here as well. This object has a very specific purpose, which is to represent a point in time for the purpose of `Participation.Process` and centralized UI presentation. Use C# types like DateTimeOffset and TimeSpan when you perform time related operations.

src-v2/NTS.Domain.Core/Objects/Start.cs
```diff
@@ -56,20 +30,24 @@ public Start(Person athlete, int number, int loopNumber, double distance, double
 
     public string StartIn()
```
achobanov: If I get this correctly this method exists to format a Timestamp as a string only for that string to be parsed back as a TimeSpan in NotTimer. Is that a good logical flow?

src-v2/NTS.Domain.Core/Objects/Startlist.cs
```diff
@@ -0,0 +1,83 @@
...
+using NTS.Domain.Core.Entities;
+
+namespace NTS.Domain.Core.Objects;
+
+public class StartList
+{
+    public StartList() { }
+
+    public List<Start> Upcoming { get; set; } = new List<Start>();
+    public List<Start> History { get; set; } = new List<Start>();
```
achobanov: We discussed that Startlist will expose collections of starts by their stage rather than having Upcoming and History.

src-v2/Not/Not.Blazor/Mud/Components/NotTrueCenterText.razor
```diff
@@ -1,10 +1,10 @@
 ﻿
 @inherits MudText
 
-<MudText Align="Align.Center" Typo="Typo" Color="Color" Class="Class" Style="@(Style + "position: absolute; left: 0; width: 100%")">
+<MudText Align="Align.Center" Typo="Typo" Color="Color" Class="@Class" Style="@(Style + "position: absolute; left: 0; width: 100%")">
```
achobanov, line 4: Instead of centering with double `MudSpacer` you're now resorting to custom css...

src-v2/Not/Not.Blazor/TM/NotSimpleTable.razor
```diff
@@ -0,0 +1,45 @@
...
+    </thead>
+    <tbody>
+        @if(Items.Count() == 0)
+        {
+            <MudText Typo="Typo.body1">@EmptyMessage</MudText>
+        }
+        @foreach (var item in Items)
+        {
+            <tr>
+                @foreach (var cellValue in item!.ToString()!.Split("|"))
```
achobanov: The idea of `Nullable` isn't to slap `!` everywhere, but to make it more obivous where you can expect null reference errors. What happens if an Item (which isn't constrained to any type at all) returns null from ToString? While returning null isn't sensible in a ToString it's completely compile-time valid and as such should be handled gracefully.

src-v2/Not/Not.Blazor/TM/NotSimpleTable.razor
```diff
@@ -0,0 +1,45 @@
...
+
+@code {
+    [Parameter, EditorRequired]
+    public IEnumerable<T> Items { get; set; } = new List<T>();
+    [Parameter, EditorRequired]
+    public string[] Headings { get; set; } = default!;
+    [Parameter]
+    public string EmptyMessage { get; set; } = "This table is still empty";
+    [Parameter]
+    public RenderFragment<T> CustomCellComponent { get; set; } = default!;
```
achobanov: CustomCell should be enough

src-v2/Not/Not.Blazor/TM/NotTimer.razor
```diff
@@ -36,7 +43,7 @@
             _timer.Enabled = false;
         }
         _displayTime = _time.ToString().Replace("-", " - ");
-        InvokeAsync(StateHasChanged);
+        await Render();
     }
 
     public void FormatTime()
```
achobanov: Is this method supposed to be public?

src-v2/Not/Not.Blazor/TM/NotTimer.razor
```diff
@@ -36,7 +43,7 @@
             _timer.Enabled = false;
         }
         _displayTime = _time.ToString().Replace("-", " - ");
```
achobanov: This formatting lgic is duplicated.

src-v2/Not/Not.Blazor/TM/NotTimer.razor
```diff
@@ -11,23 +12,29 @@
...
-        _timer = new Timer(1000);
         _timer.Elapsed += OnTimerElapsed;
         _timer.AutoReset = (_time > TimeSpan.FromMinutes(StopTimerAtTime));
         _timer.Enabled = _time > TimeSpan.FromMinutes(StopTimerAtTime);
     }
 
-    private void OnTimerElapsed(object? sender, System.Timers.ElapsedEventArgs e)
+    private async void OnTimerElapsed(object? sender, System.Timers.ElapsedEventArgs e)
     {
         _time = _time.Subtract(TimeSpan.FromSeconds(1));
```
achobanov: Why not simply use `-` operator?

achobanov: Furthermore `TimeSpan` value can be initialized once, not on every elapsed handler

src-v2/Not/Not.Blazor/TM/Pages/NotPage.razor
achobanov, line 1: I don' t like the changes on this file. It's supposed to be a generic layout solution, not tailored to specifics of it's users. What's stopping you from simply showing and hiding your relevant components within the `Main` tag in the page?

src-v2/Not/Not.Domain/DomainObject.cs
```diff
@@ -1,3 +1,10 @@
 ﻿namespace Not.Domain;
 
-public abstract record DomainObject { }
+public abstract record DomainObject
+{
+    protected string Combine(params object?[] values)
+    {
+        var sections = values.Where(x => x != null);
+        return string.Join(" | ", sections);
+    }
```
achobanov: The logic is copy-paste from `DomainEntity`....

achobanov (changes requested):

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Startlist/History.razor
StoyanGoranov: UserTabs is a collection of TabModels to render a tab for each unique PhaseNumber of the Upcoming and History collections. Idk how to use the properties without a model.

src-v2/Not/Not.Blazor/Mud/Components/NotTrueCenterText.razor
StoyanGoranov: The only change on line 4 is adding a '@' to ensure '@Class' is treated as a variable and not a hardcoded string.

src-v2/Not/Not.Blazor/TM/NotSimpleTable.razor
StoyanGoranov: The new anylizers prevent compilation without these !. Not sure how to handle them more gracefully.

src-v2/Not/Not.Blazor/TM/NotTimer.razor
StoyanGoranov: Substract seems clearer in this case.

src-v2/Not/Not.Blazor/TM/Pages/NotPage.razor
StoyanGoranov: The changes in this file allow to remove or set Elevation of the internal MudPapers and to hide the latter, if the Main or Main2 are empty. Otherwise,  if only the relevant components within the Main tags are hidden, the hardcoded paddings get displayed.

src-v2/Not/Not.Domain/DomainObject.cs
StoyanGoranov: Copied this since you have comment to use Combine in ToString(). However, as per convention Start is DomainObject since it is created during runtime and isn't saved. The Combine method was missing from DomainObject.

src-v2/NTS.Domain.Core/Objects/Startlist.cs
StoyanGoranov: Revamped Startlist

src-v2/Not/Not.Domain/DomainObject.cs
achobanov: And there isn't a possible solution to reuse code between two objects other than inheritance?

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Startlist/History.razor
achobanov: Didn't we discuss for the Startlist to expose a nexted collection of Starts by stage?

src-v2/Not/Not/Extensions/DoubleExtension.cs
achobanov: Still standing

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Startlist/History.razor
achobanov: My issue is that this is another two values model, which sounds like it's specific - TabModel - but actually has nothing to do with tabs whatsoever. I just holds 2 properties one of which is useless. Does this sound like a good design?

You can group them or put them in a dictionary.

src-v2/Not/Not.Blazor/TM/NotSimpleTable.razor
achobanov: I proposed a question for a reason: What happens if `ToString()` returns null? 

Regarding the `item` being null (the first `!`), this one is less obvious, but it's due to how Nullable is implemented in c#. The problem is that before Nullable all reference types could be assigned a null. While technically correct, this very often leads to inconsistencies which aren't super obivous for the common programmer. A good example is the `string` type, which is built-in type, but it differs from int, double, etc because it's actually a reference type - a collection of `char`s. The idea behind Nullable is to add operators such as  `?` and `!` (and another tools) to help developers make it more obvious which definitions are expected to be null and which are not. The problem is that if the changed the compilation rules, it's going to break an immense amount of existing code, which would have to be updated before migrating forward. Because of this they decided on implementing half-measures, which can help without breaking existing code. The trade off is that some stuff isn't obvious. 

In this case it leads to `item` treated as potentially nullable by the analyzers, even if `Items` is defined as `IEnummerable<T>` (not nullable). This happens because the analyzers have no idea what type `T` is. That's why it's always a good practice to constrain your generic arguments as much as possible.  In your case `Start` is a record, which is class (unless defined as record struct, in which case it's a structure , i.e. a value-type) which means you can constraint it as `@typeparam T where T : class`

src-v2/Not/Not.Blazor/TM/NotTimer.razor
achobanov: No. We're using operators when they are defined. Otherwise we should start using `5.LessThan(10)`, `var a = 5.Minus(3)` and so on. It's not clearer, it's excessively verbose.

src-v2/Not/Not.Blazor/TM/Pages/NotPage.razor
achobanov: I can see what the changes are. Revert them.

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Startlist/StartlistTable.cs
```diff
@@ -0,0 +1,35 @@
...
+using System.Threading.Tasks;
+using MudBlazor;
+using Not.Blazor.Components;
+using Not.Blazor.TM.Models;
+using Not.Services;
+using NTS.Domain.Core.Objects;
+
+namespace NTS.Judge.Blazor.Pages.Dashboard.Startlist;
+
+public abstract class StartlistTable : NotComponent
```
achobanov: This has nothing to do with a table. It could be a table if u make it inherit `NotSimpleTable` and use it as tab content. That way you'll aslo be able to reuse the filter logic.

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Startlist/StartlistTable.cs
```diff
@@ -0,0 +1,35 @@
...
+{
+    protected MudTabs _tabs = default!;
+    protected List<TabModel> _userTabs = [];
```
achobanov: You can also remove TabModel and instead have a simple collection of stages here

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Startlist/StartlistTimer.razor
```diff
@@ -8,11 +9,11 @@
 
 @code {
     [Parameter]
-    public string StartTime { get; set; } = default!;
+    public TimeSpan StartTime { get; set; } = default!;
     [Parameter]
     public double StopTimerAtTime { get; set; } = default!;
 
-    const int TimerRefreshInMilliseconds = 1000;
+    TimeSpan TimerRefreshTime = TimeSpan.FromSeconds(1);
```
achobanov: `TimeSpan` represents an interval of time, rather than a point in time, so naming it time isn't very helpful. I'd do `TimerInterval`.

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Startlist/StartlistTimer.razor
```diff
@@ -8,11 +9,11 @@
 
 @code {
     [Parameter]
-    public string StartTime { get; set; } = default!;
+    public TimeSpan StartTime { get; set; } = default!;
```
achobanov: Why would `StartTime` be a `TimeSpan`? StartTime should be a point in time, when the participant should start. `TimeSpan` represents an interval of time. It makes no sense to use it here, because now you need to know what is this interval relative to.

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Startlist/StartlistTimer.razor
```diff
@@ -22,33 +23,47 @@
...
+        }
+        else
+        {
+            var positiveSignedTime = _time.Negate();
+            var time = TimeOnly.FromTimeSpan(positiveSignedTime);
+            _displayTime = new DateTime(day, time).ToString("HH:mm:ss");
+            if(_time <= TimeSpan.FromSeconds(-1)){
+                _displayTime = " - " + _displayTime;
+            }
+        }
```
achobanov: This logic seems very obfuscated, probably because you are using a `TimeSpan` for the `StartTime` value. Also it's not formatter very well.

src-v2/Judge/NTS.Judge/Adapters/Behinds/StartlistBehind.cs
```diff
@@ -1,63 +1,45 @@
...
 using NTS.Domain.Core.Objects;
-using NTS.Domain.Objects;
+using NTS.Domain.Core.Objects.Payloads;
 using NTS.Judge.Blazor.Ports;
-using NTS.Judge.Factories;
 
 namespace NTS.Judge.Adapters.Behinds;
 
-public class StartlistBehind : ObservableBehind, IStartlistBehind
+public class StartlistBehind : ObservableBehind, IStartupInitializer, IStartlistBehind
```
achobanov: I see you've implemented in in a similar way to Handouts, but I think we can do better in retrospective, especially since we have `StartList` object already exisitng in our Domain Model:
1. Subscribe/Unsubscribe to `PhaseCompleted` in the StartList itself
2. Expose a `Changed` Event in Startlist which emits when a Start has been added
3. Subscribe to that event in `StartlistBehind` and invoke `EmitChanged`. You can even include the Action to subscribe to as a `Startlist` ctor overload to indicate that events based communication is used.

This approach will encapsulate the entirety of the business flow within the Domain which is what we're aiming to do. 

If Startlist is not created and as such not subscribed to `PhaseCompleted` it's still a non-issue, because it will be generated from existing Participations when necessary.

achobanov: If this approach works well, as I think it will, we can convert the Handouts as well.

src-v2/NTS.Domain.Core/Objects/Start.cs
```diff
@@ -10,39 +12,34 @@ public Start(
...
-        if (StartAt > Timestamp.Now())
-        {
-            return (StartAt - Timestamp.Now())!.ToString();
-        }
-        else
-        {
-            return "-" + (StartAt - Timestamp.Now())!.ToString();
-        }
-    }
+    public TimeSpan StartAt { get; private set; }
```
achobanov: Describes a point in time, rather than an interval of time.

achobanov: I can see you're still having trouble with naming and this is a good example: the object which you are working in is named `Start`, which means that simply naming this `Time` is good enough as you get the context of `Start.Time` which describes the information stored accurately. Opposingly, `History` and `Upcoming` aren't good component names, because when navigating the entire solution one is often searching by file name/class name neither `Upcoming` nor `History` provides enough information on their own, so `StartlistUpcoming` and `StartlistHistory` would be better.

src-v2/NTS.Domain.Core/Objects/Start.cs
```diff
@@ -10,39 +12,34 @@ public Start(
...
+        TotalDistance = totalDistance.RoundWholeNumberToTens();
         StartAt = startAt;
     }
 
+    public Start(Participation participation)
+    {
+        Athlete = participation.Combination.Name;
+        Number = participation.Combination.Number;
+        var phaseIndex = participation.Phases.IndexOf(participation.Phases.Current);
+        PhaseNumber = phaseIndex + 2;
```
achobanov: This is a classic case of magic string usage being hard to determine. After some pondering I've deduced that you're selecting the Phase after `Current`. Are you sure this is behaving correctly with the change you've made in `Participation.cs`? Isn't `Current` going to be the phase you need here, rather than the next one?

From a quality of code standpoint I shouldn't have to deduce things like that, it should be plain text to read. We do have a `NumberOf` extension which returns the 1-base index of an item in a collection

src-v2/NTS.Domain.Core/Objects/Startlist.cs
```diff
@@ -5,49 +5,38 @@ namespace NTS.Domain.Core.Objects;
...
+    public StartList() { }
 
-    public void OrderHistoryByAscending()
-    {
-        History = History.OrderBy(s => s.StartAt).ToList();
-    }
+    public List<Start> Starts { get; set; } = [];
 
-    public void Replace(Start start)
+    public List<Start> History
```
achobanov: Properties of the domain should not allow external actors to modify it's state unless part of the design. While the implementation of this getter currently does not allow due to the use of `ToList` (which is probably accidental, am I right) it's still not telegraphed well - it should be a `ReadOnlyCollection` or better `IReadOnlyList`.

src-v2/NTS.Domain.Core/Objects/Startlist.cs
```diff
@@ -5,49 +5,38 @@ namespace NTS.Domain.Core.Objects;
...
+        get
         {
-            Upcoming.Remove(start);
-            History.Add(start);
-            OrderHistoryByAscending();
+            var now = DateTime.Now.TimeOfDay;
+            return Starts
+                .Where(s => now - s.StartAt <= START_EXPIRY_TIME)
+                .OrderBy(s => s.StartAt)
+                .ToList();
```
achobanov: Once again, this is a copy-paste. You have to start to put more thought in reusing code. In this case, it's a textbook example, because when you change the type of `StartAt` you'll have to make changes to both properties, rather than a single method.

src-v2/NTS.Domain.Core/Objects/Startlist.cs
```diff
@@ -63,21 +52,16 @@ public void AssignStarts(IEnumerable<Participation> participations)
                         phaseIndex + 1,
                         phases[phaseIndex].Length,
                         participation.Phases.Distance,
-                        phase.StartTime!
+                        phase.StartTime!.DateTime.TimeOfDay
```
achobanov: Using `!` here is safe currently, which is OK, but ideally we should aim to avoid ! unless necessary. Because once the warning is suppressed it cannot be raised again even in valid situations. The diagnostic isn't capable enough to validate that you're filtered away null values. Instead you can do:
```
foreach (var phase in phases)
{
    if (phase.StartTime == null)
    {
        continue;
    }
    var phaseIndex = phases.IndexOf(phase);
    var start = new Start(
        participation.Combination.Name,
        participation.Combination.Number,
        phaseIndex + 1,
        phases[phaseIndex].Length,
        participation.Phases.Distance,
        phase.StartTime.DateTime.TimeOfDay
    );
    Starts.Add(start);
}
```
And avoid suppressing the warning itself.

src-v2/NTS.Domain.Core/Objects/Startlist.cs
```diff
@@ -63,21 +52,16 @@ public void AssignStarts(IEnumerable<Participation> participations)
                         phaseIndex + 1,
```
achobanov: `NumberOf` would be ideal here.

src-v2/Not/Not.Blazor/TM/NotSimpleTable.razor
```diff
@@ -22,10 +22,10 @@
                 {
                     <td>@cellValue</td>
                 }
-                @if (CustomCellComponent != default)
+                @if ( CustomCell != default)
```
achobanov: whitespace

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Startlist/StartlistTimer.razor
```diff
@@ -8,11 +9,11 @@
```
achobanov, line 11: Why not use `MudText` with `Color` instead of injecting the theme and manually selecting colors using styles?

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Startlist/StartlistTimer.razor
```diff
@@ -22,33 +23,47 @@
...
+        else
+        {
+            var positiveSignedTime = _time.Negate();
+            var time = TimeOnly.FromTimeSpan(positiveSignedTime);
+            _displayTime = new DateTime(day, time).ToString("HH:mm:ss");
+            if(_time <= TimeSpan.FromSeconds(-1)){
+                _displayTime = " - " + _displayTime;
+            }
+        }
         if (_time > TimeSpan.FromMinutes(5))
```
achobanov: These TimeSpans can also be extracted as fields

achobanov (changes requested):

src-v2/Judge/NTS.Judge/Adapters/Behinds/StartlistBehind.cs
StoyanGoranov: Changed Startlist property to private, didn't touch the rest as discussed.

src-v2/Not/Not.Domain/DomainObject.cs
StoyanGoranov: First, I thought about a common interface, but then I noticed the DomainExtension class where I can define:
```
public static string Combine(params object?[] values)
{
      var sections = values.Where(x => x != null);
      return string.Join(" | ", sections);
}
```
this would require to add the following using in all DomainEntities:
`using static NTS.Domain.Extensions.DomainExtensions;`

src-v2/NTS.Domain.Core/Objects/Start.cs
StoyanGoranov: Surprisingly, even with the change in `Participation.cs` the Current phase is not changed by `StartIfNext` it only sets the StartTime of the next phase, the Current phase is changed by SelectNext which is skipped because the Process sets the Current.IsComplete() to true, after the related isComplete variable already set in the following code:
```
        var isComplete = Current.IsComplete();
        ...
        if (isComplete && snapshot.Timestamp > Current.GetOutTime() + notProcessingWindow)
        {
            SelectNext();
        }
        return Current.Process(snapshot);
 ```
 i.e. a Phase is Completed only after the Combination finishes its previous Loop.
 
 https://github.com/Not-Endurance/not-timing-system/blob/a9bde0f632e554de2f14010f6782f47f13bf97d5/src-v2/NTS.Domain.Core/Entities/ParticipationAggregate/PhaseCollection.cs#L24

File: src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Startlist/StartlistHistory.razor [hunk 64]
```diff
@@ -0,0 +1,45 @@
...
+            </MudTabPanel>
+        }
+    </MudTabs>
+}
+
+@code{
+    [Inject]
+    public IStartlistHistory _behind { get; set; } = default!;
+
+    protected override IEnumerable<Start> _starts => _behind.History;
```
achobanov: You don't actually need this abstract property, you cans imply pass the _behind.History as argument to `CreateHeadersAndGroupByStage`.

src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Startlist/StartlistHistory.razor
```diff
@@ -0,0 +1,45 @@
...
+
+    string[] _tableHeadings = { "Number", "Athlete", "Loop", "Start Time" };
+
+    protected override async Task OnInitializedAsync()
+    {
+        await Observe(_behind);
+        CreateHeadersAndGroupByStage();
```
achobanov: I'd rename it to CreateStartlistsByStage to match with the property.

src-v2/Judge/NTS.Judge/Adapters/Behinds/StartlistBehind.cs
```diff
@@ -9,37 +9,25 @@
...
         _participationRepository = participationRepository;
-        var startlist = new StartList();
+        var startlist = new StartList(EmitChange);
         Startlist = startlist;
     }
 
-    public StartList Startlist { get; set; }
-    public List<Start> Upcoming => Startlist.Upcoming;
-    public List<Start> History => Startlist.History;
+    StartList Startlist { get; set; }
```
achobanov: Favor private fields. Use properties when
- protected, ie. has to be used internally by a child type
- public - we don't never ever want public fields
- private - in limited scenarios when they cannot be avoided, such as when using the [Inject] attribute.

src-v2/NTS.Domain.Core/Objects/Start.cs
```diff
@@ -11,40 +12,37 @@ public Start(
...
+        Time = nextPhase.StartTime!.DateTime;
     }
 
     public Person Athlete { get; private set; }
     public int Number { get; private set; }
     public int PhaseNumber { get; private set; }
     public double Distance { get; private set; }
-    public double TotalDistance { get; private set; }
-    public TimeSpan StartAt { get; private set; }
+    public DateTimeOffset Time { get; private set; }
```
achobanov: Do you need private setters?

src-v2/NTS.Domain.Core/Objects/Startlist.cs
```diff
@@ -1,37 +1,36 @@
...
 
-    public StartList() { }
+    public StartList(Action action)
+    {
+        Participation.PHASE_COMPLETED_EVENT.Subscribe(PhaseCompletedHandler);
+        ChangeHandler = action;
+    }
 
+    Action ChangeHandler { get; set; }
     public List<Start> Starts { get; set; } = [];
```
achobanov: field over property

src-v2/NTS.Domain.Core/Objects/Startlist.cs
```diff
@@ -42,26 +41,41 @@ public void AssignStarts(IEnumerable<Participation> participations)
...
+
+    TimeSpan CurrentTime()
+    {
+        return DateTimeOffset.Now.TimeOfDay;
+    }
+
+    List<Start> OrderByTimeAndPhase(IEnumerable<Start> starts)
     {
-        Starts.Add(start);
+        return starts.OrderBy(s => s.Time).OrderBy(s => s.PhaseNumber).ToList();
```
achobanov: This isn't actually going to order by time at all, because the second `OrderBy` would simply overwrite the first one. Instead you should use `OrtherBy(...).ThenBy(...)`

src-v2/NTS.Domain/Extensions/DomainExtensions.cs
```diff
@@ -9,4 +9,10 @@ public static void Update<T>(this List<T> collection, T entity)
         collection.Remove(entity);
         collection.Insert(index, entity);
     }
+
+    public static string Combine(params object?[] values)
+    {
+        var sections = values.Where(x => x != null);
+        return string.Join(" | ", sections);
+    }
```
achobanov: Seeing this makes me realize this isn't an actual extension, it's a simple static method. A more correct way to do it is to define Combine methods in both DomainEntity and DomainObject and then simply use this method, which you can put in something like `DomainObjectHelper` or `DomainModelHelper`.

src-v2/Not/Not/Formatting/FormattingHelper.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace Not.Formatting;
+
+public static class FormattingHelper
+{
+    public static string TimeSpanFormat(TimeSpan timeSpan)
```
achobanov: Simply `Format` is enough since you have the context of the type of the argument.

src-v2/Judge/NTS.Judge/Adapters/Behinds/StartlistBehind.cs
```diff
@@ -9,37 +9,25 @@
...
+    public IReadOnlyList<Start> Upcoming => Startlist.Upcoming;
+    public IReadOnlyList<Start> History => Startlist.History;
 
     protected override async Task<bool> PerformInitialization(params IEnumerable<object> arguments)
     {
         var participations = await _participationRepository.ReadAll();
         Startlist.AssignStarts(participations);
```
achobanov: Isn't it possible right now to add duplicates if it initializes twice? Why don't you remove the `AssignStarts` method, it's a bit nonsensical either way and simply pass the participations to the ctor instead? Then you can create the Startlist in the initialization.

achobanov (changes requested):

src-v2/Judge/NTS.Judge/Adapters/Behinds/StartlistBehind.cs
StoyanGoranov: Remove the AssignStarts but as consequence had to ad the following logic:
```
StartList? _startlist;
...
 public IReadOnlyList<Start> Upcoming => _startlist?.Upcoming == null ? [] : _startlist.Upcoming;
 public IReadOnlyList<Start> History => _startlist?.History == null? [] : _startlist.History;
 ```
 to avoid the error **Non-nullable field must contain a non-null value when exiting constructor. Consider declaring it as nullable.** in the ctor.

File: src-v2/Judge/NTS.Judge/Adapters/Behinds/StartlistBehind.cs (hunk 66 above)
Comment (StoyanGoranov) [reply src-v2/Judge/NTS.Judge/Adapters/Behinds/StartlistBehind.cs]: Applied the changes but adhered to private fields as you used the verb **favor**. Could you please elaborate on the clarification that private should be used only when it cannot be avoided. As it is phrased now it seems to me that we have the following options:
1. protected - use only when property is used in child type
2. private - avoid if you can
3. never declare public fields as it is probably a security vulnerability, I am not sure about the reasoning

File: src-v2/Judge/NTS.Judge/Adapters/Behinds/StartlistBehind.cs (hunk 72 above)
Comment (achobanov) [reply src-v2/Judge/NTS.Judge/Adapters/Behinds/StartlistBehind.cs]: That's right. You can do a bit better however:
```c#
public IReadOnlyList<Start> Upcoming => _startlist?.Upcoming ?? [];
```

And if you think about it it makes more sense, because you previously had an empty Startlist. What happens if you omitted to assign default values to `History` or `Upcoming`? This way is safer. That's a good illustration of the usefulness of Nullable

src-v2/Judge/NTS.Judge/Adapters/Behinds/StartlistBehind.cs
```diff
@@ -12,22 +12,21 @@ namespace NTS.Judge.Adapters.Behinds;
...
-    public IReadOnlyList<Start> History => Startlist.History;
+    public IReadOnlyList<Start> Upcoming => _startlist?.Upcoming == null ? [] : _startlist.Upcoming;
+    public IReadOnlyList<Start> History => _startlist?.History == null ? [] : _startlist.History;
 
     protected override async Task<bool> PerformInitialization(params IEnumerable<object> arguments)
     {
         var participations = await _participationRepository.ReadAll();
-        Startlist.AssignStarts(participations);
-        return Startlist.Starts.Any();
+        var startlist = new StartList(participations, EmitChange);
```
achobanov: This variable declaration is redundant.

src-v2/NTS.Domain.Core/Objects/Startlist.cs
```diff
@@ -56,16 +33,45 @@ public void AssignStarts(IEnumerable<Participation> participations)
...
-                    Starts.Add(start);
+                    _starts.Add(start);
                 }
             }
         }
+        Participation.PHASE_COMPLETED_EVENT.Subscribe(PhaseCompletedHandler);
+        ChangeHandler = action;
+    }
+
+    Action ChangeHandler { get; set; }
```
achobanov, line 44: This should also be a field.

achobanov (changes requested):

src-v2/Judge/NTS.Judge/Adapters/Behinds/StartlistBehind.cs
achobanov: I don't quite understand your question, but it seems to me that you're not making the distinction between a field and a property. Or at least you seem to be confusing them. While I think that on a lower programming level all class members are essentially fields the major distinction between a field and a property is that _fields hold state, while properties expose an API to the consumers (provide tool that can granularly configure access - getters and setters_. 

A Property has the additional advantage as being recognized by IDEs which provide useful tools for tracking their usage:
image

It is not considered safe to expose fields, because they almost always have to be modifier internally in the class, but there is no way to restrict their modification from outside of the class. With properties you can expose only a getter. The only case when fields should be exposed is when they are `const` or `static readonly` as these two combinations of modifiers ensure that the value of this field cannot be changed.

src-v2/NTS.Domain.Core/Objects/Startlist.cs
achobanov: Please not that when I select a multi-lined comment usually my comment applies to both lines, as you only converted the `Starts` to a field.

achobanov (approved):

achobanov:
Merging without resolving because I'm in a hurry and current unresolved are only explanations, all requested changes are impelemented

PR #302, opened 2024-10-03
Fix duplicate state with serialization

### LINKED ISSUE #301: Fix state duplication
## The Problem 
- Either the domain entities are restricted not to reference other entities in 1-to-1 relationships
- or the resulting json contains duplicate information.

### Domain perspective
It makes everything more cumbersome for the domain if you start using Ids instead of directly referencing objects as domain models can con longer encapsulate all information necessary for their operation and it has to be provisioned separately. 

### Persistence perspective
References **have to be resolved**, otherwise updating an unresolved reference may result in the parent object containing obsolete information: If a `Phase` has a `Loop` and `Loop.Length` is updated, `Phase.Loop.Length` will not updated and it's not stuck with wrong data.

## Solution
Do the leg work in the repositories. Unfortunately that does come with notable drawbacks:
-Repositories now have to implement actual update logic (previously the object was simply replaced, but that no longer works, because if a `Loop` is removed during update the reference to `Phase` breaks and they are serialized as separate objects)
- Domain encapsulation is weakened, as in order to update an object instance the repo has to be able to set the object's properties. The lesser evil is to declare the setters as `internal` and use `InternalsVisibleToAttribute` in order to restrict outside intervention to only within the domain and persistence projects.

achobanov:
closes #301


src-v2/Judge/NTS.Judge.Blazor/Pages/Dashboard/Ranklist/RanklistRow.razor
```diff
@@ -75,24 +75,10 @@
...
 
     protected override void OnParametersSet()
     {
-        var participation = _behind.Get(Entry.ParticipationId);
-        if (participation == null)
-        {
-            return;
-        }
-
+        var participation = Entry.Participation;
```
StoyanGoranov, line 81: 👀

src-v2/NTS.Domain.Core/Objects/Regional/FeiRanker.cs
```diff
@@ -1,36 +1,26 @@
...
 
-    protected IOrderedEnumerable<Pair> OrderByNotEliminatedAndRanked(
-        IEnumerable<RankingEntry> entries,
-        IEnumerable<Participation> participations)
+    protected IOrderedEnumerable<RankingEntry> OrderByNotEliminatedAndRanked(IEnumerable<RankingEntry> entries)
     {
-        var ids = entries.Select(x => x.ParticipationId).ToList();
         return entries
-            .Select(x => new Pair(x, participations.First(y => x.ParticipationId == y.Id)))
             .OrderBy(x => x.Participation.IsNotQualified())
```
StoyanGoranov, line 17: 👀

src-v2/NTS.Domain.Core/Objects/Regional/FeiRanker.cs
```diff
@@ -1,36 +1,26 @@
...
-            .ThenByDescending(x => !x.RankingEntry.IsRanked);
+            .ThenByDescending(x => !x.IsRanked);
```
StoyanGoranov: 👀

StoyanGoranov (approved):

PR #316, opened 2024-10-27
fix Rfid unexpected explosions

StoyanGoranov:


achobanov:
I suggest we close this, as it's indent was to hotfix the situation. Instead you can start working on: https://github.com/Not-Endurance/not-timing-system/issues/318 as a long-term solution once you get the RFID gear.

StoyanGoranov:
abort mission, abort mission

PR #337, opened 2024-12-02
asenovgrad fixes

### LINKED ISSUE #319: Asenovgrad fixes
## Checklist
- [x] Add `CompetitionType.Championship`  - Don't use recovery-based ranking when type is Championship in BulgariaRanker
- [x] Add `AthleteCategory.Companion` - companions are Seniors who accompany children and do not compete. Having a separate category excludes the Companion from the Children Ranklist. Furthermore speed restrictions aren't necessary for Companions
- [x] Prevent creation of empty rankings. Once done you can revert [this commit](https://github.com/Not-Endurance/not-timing-system/commit/fbe385b4d52a9bee34913297d71725feb1c80d14)
- [x] Pass Setup EnduranceEvent.Id to Core EnduranceEvent.Id when starting event. This is necessary because Witness applications keep their data based on the Core.EnduranceEvent.Id. As of now when we delete CoreState and start anew, we generate a new ID and Witness discards their current state, which is why we lost the data at the start
- [x] Familiarize with the [logging infrastructure](https://github.com/Not-Endurance/not-timing-system/tree/develop/src-v2/Not/Not/Logging) and log non-validation exceptions in `SafeHelper`. Logging isn't currently used, so there may be implementational problems
- [x] Prevent `JuniorOrYoundAdult` Athletes in Championship competitions. Indicate that those athletes should be configured as `Seniors` instead

StoyanGoranov:
Closes: https://github.com/Not-Endurance/not-timing-system/issues/319


src/Judge/NTS.Judge/Core/Behinds/CompetitionParentContext.cs
```diff
@@ -47,16 +49,32 @@ public void Remove(Phase child)
...
+        ValidateAthleteCategory(child);
         Entity!.Update(child);
     }
 
     public void Remove(Participation child)
     {
         Entity!.Remove(child);
     }
+
+    void ValidateAthleteCategory(Participation child)
```
achobanov: Is this the correct place to validate?

src/Judge/NTS.Judge/Core/Behinds/CompetitionParentContext.cs
```diff
@@ -47,16 +49,32 @@ public void Remove(Phase child)
...
+    {
+        if (
+            child.Combination.Athlete.Category == AthleteCategory.JuniorOrYoungAdult
+            && Entity!.Type == CompetitionType.Championship
+        )
+        {
+            throw new DomainException(
+                nameof(Participation.Combination),
+                "Athletes participating in Championship Competitions must be of Senior category."
```
achobanov: This message isn't correct. What about Chidlren?

achobanov (changes requested):

achobanov (approved):

PR #345, opened 2024-12-20
fix SignalR two-way communication

### LINKED ISSUE #336: Fix SignalR two-way communication
## Overivew
Currently SignalR's Hub -> Client communication does not work when the hub isn't started using `Process.Start`. Currently the workaround is to effectively long poll from the client, which is very limiting. The simplest way to solve this as of now is segregate the Hub (`NTS.Judge.MAUI.Server currently`), start it from Judge using `Process.Start` and then imlement a SignalR connection between Judge and Hub as well. 

## Examples

Lets illustrate this wilth one example for each vector

### Judge > Witness vector
When `PhaseCompleted` event is raised Judge should send a new Start entry to Witness apps:
1. Judge invokes `IEmsStartlistClientProcedures.ReceiveEntry`
2. Hub receives `ReceiveEntry` invocation and _relays_ by invoking the same procedure on its clients (Judge's client should not register this procedure, because it doesn't need it)
3. Witness receives `ReceiveEntry` same as before, no changes on Witness side

In order to do this you will need to create a StartlistClient in Judge, similar to the one in [Witness](https://github.com/Not-Endurance/not-timing-system/blob/develop/src/Witness/EMS.Witness/Rpc/StartlistClient.cs). Remember - because Judge is only _acting like a hub, but is technically a client_ `ReceiveEntry` should be registered invoked (L:29), rather than registered (L:15).

Once this example is working you should have some familiarity with how the communication operates on a high level. Now let's do better:
- instead of using `IEmsStartlistClientProcedures` in Judge defina a new interface using NTS entities directly. Design `ReceiveEntry` procedure's equivalent run it by me and implement it.
- encapsulate the Anti-corruption layer (ACL) logic enirely in the Hub project and execute it when an invocation is received from Judge

### Witness > Judge vector
When Witness connects it has to load all upcoming Starts:
1. WItness invoke `InvokeHubProcedure.SendStartlist` hub proceduce
2. Hub receives `SendStartlist` invocation
    We have a problem, because Hub is now separate application - i.e. it no longer has access to the Judge dependency container and thus cannot directly access the database. It shouldn't be too difficult to get that access though - we can mimic how Judge's repositories are configured ( we need to configure a `JsonFileStore` and create a repository ) and now we have access to the database. The Hub **should have readonly** access to prevent concurrency-related problems with Judge. Once we have that setup we can now implement `IStartlistHubProcedures.SendStartlist` and send the data to Witness (again Witness shoudln't need to change for this)
    
The Judge is now completely excluded from this process. 

### The rest
The other procedures should follow suite. By the end we'll have a Hub who's implementation is very losely coupled with Judge and having effectively two types of clients: Judge and one or more Witness. 

## Checklist
- [x] Implement `JudgeClient` by inherting from `RpcClient` and implement NTS-grade procedures. **This code should not be treated as legacy**
- [x] Visual que for connection status in Judge (see how it's implemented in Witness) - this one shuold be more polished
- [x] Implement `JudgeHubProcedures` and `WitnessHubProcedures` in order to achieve the same features as current EMS.
- [x] Implement exception handling and logging in the Hub (make sure that hub's log file does not conflic with Judge's log file)
- [x] Dispose of the server when the Judge application closes or crashes
- [x] Rename the SignalR server and move it to another `Relay` folder

**All RpcHub clients should reuse a single SignalRSocket connection**. There is no downside to this that I'm aware of ATM, except that you're limited to a single Hub class. Defining any more would require a separate connection which we don't want at the moment, because it makes debugging connection-related problems harder.

StoyanGoranov:
Closes: https://github.com/Not-Endurance/not-timing-system/issues/336


src/NTS.Application/RPC/IWitnessClientProcedures.cs
```diff
@@ -1,5 +1,9 @@
-﻿namespace NTS.Application.RPC;
+﻿using NTS.ACL.RPC.Procedures;
 
-public interface IWitnessClientProcedures
-{
-}
+namespace NTS.Application.RPC;
+
+public interface IWitnessClientProcedures
```
achobanov: The idea of this interface was to eventually house non-legacy methods. We don't need it atm, so just rename it to ILegacyWitnessClientProcedures

src/NTS.Application/RPC/IWitnessHubProcedures.cs
```diff
@@ -1,5 +1,8 @@
-﻿namespace NTS.Application.RPC;
+﻿using NTS.ACL.RPC.Procedures;
 
-internal interface IWitnessHubProcedures
-{
-}
+namespace NTS.Application.RPC;
+
+public interface IWitnessHubProcedures
```
achobanov: Same thing - since we cannot encapsulate the difference between legacy and non-legacy Witness (the signatures change) there is no beneft in naming this interface as non-legacy. Just rename to `ILegacyWitnessHubProcedures`

src/Judge/NTS.Judge.MAUI.Server/RPC/Procedures/IClientProcedures.cs
```diff
@@ -1,7 +1,5 @@
-﻿using NTS.ACL.RPC.Procedures;
+﻿namespace NTS.Judge.MAUI.Server.RPC.Procedures;
 
-namespace NTS.Judge.MAUI.Server.RPC.Procedures;
-
-public interface IClientProcedures : IEmsClientProcedures
+public interface IClientProcedures
```
achobanov: THis interface seems unnecessary

achobanov (changes requested):

src/Judge/NTS.Judge/Core/Behinds/RemoteConnectionsCount.cs
```diff
@@ -0,0 +1,23 @@
+﻿using Not.Application.RPC;
+
+namespace NTS.Judge.Core.Behinds;
+
+public class RemoteConnectionsCount : IConnectionsCounter
```
achobanov: What kind of "component" is this?

src/Judge/NTS.Judge/Core/Behinds/RemoteConnectionsCount.cs
```diff
@@ -0,0 +1,23 @@
...
+{
+    public List<string> ActiveConnections { get; } = [];
+
+    public void AddConnection(string connectionId)
+    {
+        ActiveConnections.Add(connectionId);
+    }
```
achobanov: Why expose public List and public methods to modify the list?

achobanov: What's the problem of exposing excessive properties?

achobanov: Hint: _encapsulation_. How can we improve encapsulation?

src/Judge/NTS.Judge/RPC/JudgeRpcClient.cs
```diff
@@ -1,16 +1,56 @@
...
+        }
+    }
+
+    public Task IncrementConnectionCount(string connectionId)
+    {
+        _remoteConnections.AddConnection(connectionId);
+        return Task.CompletedTask;
+    }
+
+    public Task DecrementConnectionCount(string connectionId)
```
achobanov: Doesn't follow the `Receive`, `Send`, `Get` convention. Also what connection?

achobanov (changes requested):

src/Judge/NTS.Judge/Core/Behinds/RemoteConnectionsCount.cs
StoyanGoranov: Encapsulation can be improved by making the stateful property private.

src/Judge/NTS.Judge.Blazor/Shared/Components/ConnectionStatus/ConnectionStatus.razor.cs
```diff
@@ -0,0 +1,49 @@
...
+
+    [Inject]
+    IConnectionsCounter ConnectionsCount { get; set; } = default!;
+
+    protected override async Task OnInitializedAsync()
+    {
+        await Observe(ConnectionsCount);
+        _isConnected = RpcSocket.IsConnected;
+        RpcSocket.Error += HandleRpcErrors;
+        RpcSocket.ServerConnectionChanged += HandleServerConnectionChanged;
```
achobanov: If the component subscribes to events, it should also take care to unsubscribe when being disposed. 

Additionally, is this the best place for this logic. Does this UI component need to know about SignalR, RPCs and connections?

src/Judge/NTS.Judge.MAUI/App.xaml.cs
```diff
@@ -11,7 +11,7 @@ public App(IEnumerable<IStartupInitializer> initializers)
 
         MainPage = new MainPage();
 
-        ContextHelper.SetApplicationName("nts");
+        ContextHelper.ConfigureApplicationName("nts");
```
achobanov: random change?

src/Judge/NTS.Judge.MAUI/MauiProgram.cs
```diff
@@ -31,20 +31,15 @@ static void StartHub()
     {
         try
         {
+            var parentPid = Process.GetCurrentProcess().Id;
             var currentDirectory = Directory.GetCurrentDirectory();
             var info = new ProcessStartInfo
             {
-                FileName = Path.Combine(currentDirectory, "NTS.Judge.MAUI.Server.exe"),
+                FileName = Path.Combine(currentDirectory, "NTS.SignalR.Server.exe"),
```
achobanov: Magic string should be a constant instead. We have constant files for most projects.

src/Judge/NTS.Judge/Core/Behinds/Adapters/RemoteConnectionsBehind.cs
```diff
@@ -0,0 +1,33 @@
+﻿using Not.Application.Behinds.Adapters;
+using Not.Application.RPC;
+using Not.Blazor.Ports;
+
+namespace NTS.Judge.Core.Behinds.Adapters;
+
+public class RemoteConnectionsBehind : ObservableBehind, IConnectionsCounter
+{
+    List<string> ActiveConnections { get; } = [];
```
achobanov: Are there any other type of collections?

Private members should be fields by default and properties if only necessary.

src/Judge/NTS.Judge/Core/Behinds/Adapters/RemoteConnectionsBehind.cs
```diff
@@ -0,0 +1,33 @@
...
+
+    protected override Task<bool> PerformInitialization(params IEnumerable<object> arguments)
+    {
+        var result = ActiveConnections.Any();
+        return Task.FromResult(result);
+    }
+
+    public void AddConnection(string connectionId)
```
achobanov: Connection seems redundant in the name

src/Judge/NTS.Judge/Core/Behinds/Adapters/RemoteConnectionsBehind.cs
```diff
@@ -0,0 +1,33 @@
...
+        EmitChange();
+    }
+
+    public void RemoveConnection(string connectionId)
+    {
+        ActiveConnections.Remove(connectionId);
+        EmitChange();
+    }
+
+    public int GetConnectionsCount()
```
achobanov: You could expose a Readonly list instead of providing custom definitions for built in functionalities such as Count.

src/Judge/NTS.Judge/RPC/JudgeRpcClient.cs
```diff
@@ -1,16 +1,70 @@
...
 using NTS.Application.RPC;
+using NTS.Domain.Core.Aggregates;
 using NTS.Domain.Core.Objects.Payloads;
+using NTS.Domain.Objects;
+using NTS.Judge.Core;
 
 namespace NTS.Judge.RPC;
 
-public class JudgeRpcClient : RpcClient, IJudgeRpcClient
+public class JudgeRpcClient : RpcClient, IJudgeRpcClient, IStartupInitializer
```
achobanov: Clients can be separate, as long as they reuse the same IRpcSocket instance - it's still one connection. You can do `ParticipationClient` and `SateliteConnectionsClient` for example.

src/NTS.ACL/Handshake/JudgeHandshakeService.cs
```diff
@@ -3,7 +3,7 @@
 using System.Text;
 using Not.Injection;
 using NTS.ACL.Handshake;
-using static NTS.Judge.MAUI.Server.ACL.Constants;
+using static NTS.SignalR.Server.ACL.Constants;
```
achobanov: Namespaces should reflect the relative path from the solution root directory.

src/NTS.Application/ApplicationConstants.cs
```diff
@@ -1,4 +1,4 @@
-﻿using NTS.Judge.MAUI.Server.ACL;
+﻿using NTS.SignalR.Server.ACL;
```
achobanov: Namespace

src/Not/Not.Application/RPC/IConnectionsCounter.cs
```diff
@@ -0,0 +1,11 @@
+﻿using Not.Blazor.Ports;
+using Not.Injection;
+
+namespace Not.Application.RPC;
+
+public interface IConnectionsCounter : IObservableBehind
```
achobanov: Is counter a good name for it based on the functionality it exposes?

src/Not/Not/Contexts/ContextHelper.cs
```diff
@@ -4,22 +4,10 @@ namespace Not.Contexts;
...
-
-        var basePath =
-#if DEBUG
-            $"C:\\tmp\\{_applicationName}";
-#else
-        Directory.GetCurrentDirectory();
-#endif
-        return Path.Combine(basePath, subdirectory);
+        _applicationName = applicationName;
+        return _applicationName;
```
achobanov: ?

src/Not/Not/Filesystem/FileContextHelper.cs
```diff
@@ -15,7 +15,7 @@ public static void SetDebugRootDirectory(string applicatioName)
         _applicationName = applicatioName;
     }
 
-    public static string GetAppDirectory(string subdirectory)
+    public static string GetAppDirectory(string subdirectory = "")
```
achobanov: Why is this necessary?

src/Not/Not/ProcessUtility/ProcessServiceContext.cs
```diff
@@ -0,0 +1,11 @@
+﻿namespace Not.ProcessUtility;
+
+public class ProcessServiceContext
+{
+    public ProcessServiceContext(string parentPID)
+    {
+        ParentProcessID = int.Parse(parentPID);
+    }
+
+    public int ParentProcessID { get; set; }
```
achobanov: Are consumers supposed to be able to set this value? 

Additionally we don't CAPS abbreviations in dotnet members. I.e. `ParentProcessId`

src/Not/Not/ProcessUtility/ProcessService.cs
```diff
@@ -0,0 +1,33 @@
+﻿using Microsoft.Extensions.Hosting;
+
+namespace Not.ProcessUtility;
+
+public class ProcessService : BackgroundService
```
achobanov, line 5: One reason this being a HostedService isn't an ideal solution is that this may be problematic or at least pointless if the SignalR.Server starts on its own.

src/Relay/NTS.SignalR.Server/Middleware/ExceptionHandlingHubFilter.cs
```diff
@@ -0,0 +1,64 @@
...
+
+    public Task OnConnectedAsync(HubLifetimeContext context, Func<HubLifetimeContext, Task> next)
+    {
+        try
+        {
+            return next(context);
+        }
+        catch (HubException ex)
+        {
+            HandleHubException(ex, "OnConnectedAsync");
```
achobanov: Use nameof instead of string literals.

src/Relay/NTS.SignalR.Server/Middleware/ExceptionHandlingHubFilter.cs
```diff
@@ -0,0 +1,64 @@
...
+    {
+        Console.WriteLine($"Calling hub method '{invocationContext.HubMethodName}'");
+        try
+        {
+            return await next(invocationContext);
+        }
+        catch (HubException ex)
+        {
+            HandleHubException(ex, invocationContext.HubMethodName);
+            return null;
```
achobanov: Why null?

src/Relay/NTS.SignalR.Server/Middleware/ExceptionHandlingHubFilter.cs
```diff
@@ -0,0 +1,64 @@
...
+    public Task OnConnectedAsync(HubLifetimeContext context, Func<HubLifetimeContext, Task> next)
+    {
+        try
+        {
+            return next(context);
+        }
+        catch (HubException ex)
+        {
+            HandleHubException(ex, "OnConnectedAsync");
+            return Task.FromException(ex);
```
achobanov: Why not use await above?

src/Relay/NTS.SignalR.Server/Middleware/ExceptionHandlingHubFilter.cs
```diff
@@ -0,0 +1,64 @@
+﻿using Microsoft.AspNetCore.SignalR;
+using Not.Logging;
+using Not.Notify;
+
+namespace NTS.SignalR.Server.Middleware;
```
achobanov: Also namespace doesn't match location. Since we're naming it Relay, then the project can simply be NTS.Relay, as it is a single layer applicaiton

src/Relay/NTS.SignalR.Server/NTS.SignalR.Server.csproj
```diff
@@ -14,9 +14,9 @@
   </ItemGroup>
 
   <ItemGroup>
-    <ProjectReference Include="..\..\NTS.ACL\NTS.ACL.csproj" />
-    <ProjectReference Include="..\..\NTS.Application\NTS.Application.csproj" />
-    <ProjectReference Include="..\..\NTS.Storage\NTS.Storage.csproj" />
+      <ProjectReference Include="..\..\NTS.Application\NTS.Application.csproj" />
+      <ProjectReference Include="..\..\NTS.Storage\NTS.Storage.csproj" />
+	  <ProjectReference Include="..\..\NTS.ACL\NTS.ACL.csproj" />
```
achobanov: indentation is messy

src/Relay/NTS.SignalR.Server/Program.cs
```diff
@@ -0,0 +1,45 @@
...
+    });
+    builder.Services.AddHostedService<ProcessService>();
+}
+
+builder
+    .ConfigureLogging()
+    .AddFilesystemLogger(logFileConfig =>
+    {
+        logFileConfig.Path = FileContextHelper.GetAppDirectory();
+        logFileConfig.Name = ContextHelper.ConfigureApplicationName("NTS.Judge.Server");
```
achobanov: Name can be dynamic here - you can get the current type's namespace for example.

achobanov (changes requested):

src/Not/Not/Contexts/ContextHelper.cs
StoyanGoranov: `ContextHelper` was up for deletion as its logic was duplicated in `FileContextHelper` and only used at `App.xaml.cs`. Alternatively, we can delete `ContextHelper` and replace it with `FileContextHelper`.

src/Judge/NTS.Judge.MAUI/App.xaml.cs
StoyanGoranov: `ContextHelper` was up for deletion as its logic was duplicated in `FileContextHelper` and only used at `App.xaml.cs`. Alternatively, we can delete `ContextHelper` and replace it with `FileContextHelper`.

src/Not/Not.Application/RPC/IConnectionsCounter.cs
StoyanGoranov: Naming it Behind is probably better since it is used as one.

src/Judge/NTS.Judge/Core/Behinds/Adapters/RemoteConnectionsBehind.cs
StoyanGoranov: Unsure whether you mean to use another type of collection or that there aren't InactiveConnections. Therefore addressed both by using the `HashSet` type and renaming the field to `_connections`

src/Not/Not/Filesystem/FileContextHelper.cs
StoyanGoranov: So that we don't need a separate overload for method calls when subdirectory is not going to be created.

src/Not/Not/ProcessUtility/ProcessService.cs
StoyanGoranov: This is addressed in the following conditional statement:
```
if (args.Length > 0)
{
    builder.Services.AddSingleton<ProcessServiceContext>(provider =>
    {
        var parentProcessID = args[0];
        return new ProcessServiceContext(parentProcessID);
    });
    builder.Services.AddHostedService<ProcessService>();
}
```
at Program.cs of the server where the HostedService is registered only if the server is started with the expected argument array.

src/Relay/NTS.SignalR.Server/Middleware/ExceptionHandlingHubFilter.cs
StoyanGoranov: Because the `InvokeMethodAsync` of the filter must return an object which represents the next method to be invoked. Not sure if there is a point in making the handler return a task just so it can be awaited. What do you think?

StoyanGoranov: Also not sure if returning `Task.FromException` is good either.

src/Relay/NTS.SignalR.Server/NTS.SignalR.Server.csproj
StoyanGoranov: this seems like some git mixstake

src/Judge/NTS.Judge.Blazor/Shared/Components/ConnectionStatus/ConnectionStatus.razor.cs
StoyanGoranov: https://github.com/Not-Endurance/not-timing-system/pull/345/commits/e32e03f0c4bc2a00a31519cb7fceb5cd28eb82b8

src/Judge/NTS.Judge/RPC/JudgeRpcClient.cs
StoyanGoranov: https://github.com/Not-Endurance/not-timing-system/pull/345/commits/46465036366aeced98b1c8b1e9585ead05a29ee9

src/Judge/NTS.Judge.Blazor/Shared/Components/ConnectionStatus/ConnectionStatus.razor.cs
achobanov: Segregate IConnectionsBehind interface

src/Judge/NTS.Judge.MAUI/App.xaml.cs
achobanov: Use `typeof(x),Namespace!` instead of secondary overload and delete, transfer functinality to FileContextHelper and  ContextHelper entirely

src/Not/Not/Filesystem/FileContextHelper.cs
achobanov: But we want subdirectories in order to be consistent

src/Not/Not/ProcessUtility/ProcessService.cs
achobanov: Check agains concrete argument in args

src/Relay/NTS.SignalR.Server/Middleware/ExceptionHandlingHubFilter.cs
achobanov: Simply use async when not returning

achobanov (changes requested):
few comments left

achobanov (approved):

Issue #356, opened 2025-01-19
Integrate Judge import with Nexus


achobanov:
## Overview
Judge should seemlessly integrate with Nexus and import results with a single click.

## Features
We import in 3 separate collections:
1. Horses - probably reverse? separate issue?
2. Athletes - probably reverse? separate issue?
3. Ranklists - ensure `new Ranklist` is used before import otherwise Rank property is not populated. Probably redesign

## Implementation
Create Nexus HTTP client that uses the write functions to import data

PR #370, opened 2025-03-01
NTS Nexus Functions

### LINKED ISSUE #353: Implement Nexus web functions
## Overview
Nexus stores NTS data in the long term. It needs the following capabilities:
- Endpoint to upload Judge results
- Endpoint to insert RanklistEntry records manually
- Endpoint to query the current database

## Implementation
We need NTS grade implementation with adjusted project structure, new Domain project (if needed) and other conventions

## Domain
Figure out how to domain model -> Mayhaps create a "Mapper domain" as it needs to:
1. validate the input and map it to MongoDB documents
2. Map from documents to correct Domain models. Figure out how to handle cross-boundary stuff - Each domain may be made aware of the storage, which isn't ideal. Alternatively we could have domains being aware of other domains which is also not ideal.

achobanov:
closes #353


achobanov:
Also implements #356

PR #400, opened 2025-03-21
fix PhaseUpdateModel FromEntity applying UTC time

StoyanGoranov:
fix: stop adding the diff between UTC time and Local time with ToDateTimeOffset method calls

behavior occurred because DateTimeOffset returns UTC time, UTC value applied to inputs via PhaseUpdateModel FromEntity and UTC time was being set, resulting in Phase time getting offset with each update


src/NTS.Domain/Objects/Timestamp.cs
```diff
@@ -116,7 +116,7 @@ public int CompareTo(Timestamp? other)
 
     public DateTimeOffset ToDateTimeOffset()
     {
-        return _stamp; //TODO: test reference modification
+        return _stamp.LocalDateTime;
```
achobanov, line 114: This doesn't seem correct even if it is right

achobanov (approved):

Issue #364, opened 2025-01-25
Contract: Improve SignalR connectivity feedback


achobanov:
## Overview
- [x] Improve notification feedback with more user-friendly messages. (Witness)
- [x] Reduce duplication in notifications in Witness (currently it spams too much and with not very helpful messages) (Witness)
- [x] During local testing I saw the Judge status component spinning while having established connection. I think we should change our approach and try using Timers are you previously attempted, which run on every second and update the status of the spinner. (Judge) - **Resolution: turned out the property used in the UI wasn't initialized as part of the ConnectionBehind state**
- [x] User friendly notification message if Judge/Witness cannot handshake (i.e. Relay is down)  (Witness)
       This error means user is probably connected to wrong Wi-fi network or is too far from the router to connect
- [x] User friendly message if SignalR connection is lost (Witness)
       This error means handshaking is successful (device IP is obtained) 
- [x] **Battle test this**. I.e. setup a small event with 5-10 users and test end-to-end until it is confirmed to behave correctly. 

_Note: Witness should still allow users to snapshot without active connection_

### SignalR
See implementation in: [SignalRScoket](https://github.com/Not-Endurance/not-timing-system/blob/develop/src/Not/Not.Application/RPC/SignalR/SignalRSocket.cs)

### Witness
Usage in witness is in [NavMenu.razor](https://github.com/Not-Endurance/endurance-management-system/blob/release-witness-v5/src/Witness/EMS.Witness/Shared/NavMenu.razor) component 

## Payment
650 BGN

PR #404, opened 2025-03-25
Improve SignalR connectivity feedback

StoyanGoranov:
Related to: https://github.com/Not-Endurance/not-timing-system/issues/364 
but doesn't close it since issue also includes changes to EMS.


src/Judge/NTS.Judge/Core/Behinds/Adapters/ConnectionsBehind.cs
```diff
@@ -24,6 +24,8 @@ public ConnectionsBehind(IRpcSocket rpcSocket)
 
     protected override Task<bool> PerformInitialization(params IEnumerable<object> arguments)
     {
+        ServerConnectionStatus = RpcConnectionStatus.Disconnected;
+        IsServerConnected = false;
```
achobanov: I don't think these lines are necessary. Disconnected and false are both default values for their types. Which is not consistent actually with how we do enums - they should not contain 0 as value. But keep RpcConnectionStatus as is for now, we'll revisit once we do the NTS Witness.

achobanov: Furthermore is `IsServerConnected` simply checking the connection status then make it readonly:
```c#
public bool IsServerConnected => ServerConnectionStatus == RpcConnectionStatus.Connected;
```

StoyanGoranov: Initializing the `ServerConnectionStatus` and `IsServerConnected` properties in `PerformInitialization` ensures they are part of the Behind state. This is needed because the properties are used in the `ConnectionStatus` component and changes to the status of the connection are reflected by the UI this way. 

This effectively eliminates the need to introduce a Timer which checks the status of the connection every second.

StoyanGoranov: Setting the IsServerConnected value is needed to ensure it is part of the `ConnectionsBehind` state.

achobanov: Remove it and you'll see it will still be working.

PR #415, opened 2025-04-04
Add Ranklist Protocol header

### LINKED ISSUE #378: Finalize static ranking protocols
## Overview
Work with Teddy to finalize the protocol

## Payment
350 BGN

StoyanGoranov:
Closes: https://github.com/Not-Endurance/not-timing-system/issues/378


src/Judge/NTS.Judge.Blazor/Core/ParticipationTable.razor.cs
```diff
@@ -22,20 +22,24 @@ public partial class ParticipationTable
     [Parameter]
     public PhaseCollection? Phases { get; set; }
 
+    [Parameter]
+    public bool Vertical { get; set; } = default;
```
achobanov: `Vertinal` is not a good Bool name

src/Judge/NTS.Judge.Blazor/Core/Rankings/IRankingBehind.cs
```diff
@@ -1,12 +1,15 @@
...
 using NTS.Domain.Core.Aggregates;
 using NTS.Domain.Core.Objects;
+using NTS.Domain.Core.Objects.Documents;
 
 namespace NTS.Judge.Blazor.Core.Rankings;
 
 public interface IRankingBehind : IObservableBehind, ISingleton
 {
+    RanklistDocument? Document { get; }
     Ranklist? Ranklist { get; }
```
achobanov: Still don't like these two being exposed in a single interface. If unavoidable we should probably separate the interfaces.

src/Judge/NTS.Judge.Blazor/Core/Rankings/Protocols/ProtocolRow.razor
```diff
@@ -3,48 +3,71 @@
...
+        </MudStack>
+    }
+    else
+    {
+        <MudStack Justify="Justify.Center" AlignItems="AlignItems.Center" Style="width:140px;">
+            <MudText Typo="Typo.body2" Align="Align.Center" Class="mud-button-text">
+                @Competition_ongoing_string
+            </MudText>
+        </MudStack>
+    }
```
achobanov, line 72: Is this section really necessary? IMO it should be: If eliminated : else

src/Judge/NTS.Judge.Blazor/Core/Rankings/Protocols/ProtocolRow.razor.cs
```diff
@@ -15,7 +15,14 @@ public partial class ProtocolRow
...
     {
-        _rankText = Entry.Rank?.ToString() ?? Incomplete_string;
+        if (!Entry.Participation.IsEliminated())
+        {
+            _rankText = Entry.Rank?.ToString() ?? Incomplete_string;
+        }
+        else
+        {
+            _rankText = " ";
+        }
```
achobanov: Cant you simply keep this null?

src/Judge/NTS.Judge.MAUI/wwwroot/images/logo-TEF.png
achobanov, line 1: Can Remove this one as we probably won't be needing it.

src/Judge/NTS.Judge.MAUI/wwwroot/images/logo-nak.jpg
achobanov, line 1: Perhaps leave commented hrefs for every image in the protocol razor file. Otherwise we'll probably miss it if we need it again.

src/Judge/NTS.Judge/Core/Behinds/Adapters/RanklistBehind.cs
```diff
@@ -50,6 +52,15 @@ protected override async Task<bool> PerformInitialization(params IEnumerable<obj
...
         Ranklist = new Ranklist(ranking);
+        var enduranceEvent = await _events.Read(0);
+        if (enduranceEvent == null)
+        {
+            NotifyHelper.Warn("Event is not started yet");
+            return false;
+        }
+        var officials = await _officials.ReadAll();
+        var ranklistDocument = new RanklistDocument(Ranklist, enduranceEvent, officials);
+        Document = ranklistDocument;
```
achobanov: I am thinking that we don't need to duplicate this here and in the Select method below. Can we instead expose CreateDocument method which will return the document from the currently selected Ranklist, for example

src/Not/Not.Blazor/Components/NPrintableComponent.cs
```diff
@@ -29,6 +31,7 @@ protected async Task OpenPrintDialog()
         InvokeToggle();
         await PrintInterop.OpenPrintDialog();
         InvokeToggle();
+        //PrintingClassName = string.Empty;
```
achobanov: Remove

achobanov (changes requested):

src/Judge/NTS.Judge.Blazor/Core/Rankings/Protocols/ProtocolRow.razor.cs
StoyanGoranov: I can but if ```_rankText``` is null the top most MudText component shouldn't be displayed and the whole column for unranked participants get misaligned to the top. 
Alternatively, I can add a condition in the MudText, if ```_rankText == null``` display empty string. Should I use this approach instead?

src/Judge/NTS.Judge.Blazor/Core/Rankings/Protocols/ProtocolRow.razor
StoyanGoranov: Participation has 3 states - Ongoing, Eliminated and Completed. I think it is best these states are represented in the Total column of the Protocol to let the users know why it isn't populated. Also avoiding some accidental misalignments of content when empty.

src/Judge/NTS.Judge.Blazor/Core/Rankings/IRankingBehind.cs
StoyanGoranov: Not that it is unavoidable, just don't know whether it makes sense to call ```Document.Ranklist``` in Components that don't have anything to do with printing/exporting. Hence splitting might be best.

src/Judge/NTS.Judge.Blazor/Core/ParticipationTable.razor.cs
StoyanGoranov: Changed to IsVertical although is seems kind of weird for a component parameter.

achobanov: `IsVerticalAligned`

src/Judge/NTS.Judge.Blazor/Core/Rankings/IRankingBehind.cs
achobanov: Of course it doesn't make sense. Instead you could only create the Document where it makes sence. For example `GetDocument` method

src/Judge/NTS.Judge.Blazor/Core/Rankings/Protocols/ProtocolRow.razor
achobanov: Well then there is a better way to do it. The Domain Model's job is to reflect the business logic as close as possible. Insteasd of having contrived checks that relay no informaiton at all, you should extend the Domain to match the use-case and end up with something like:
```
if Eliminted
   ...
else if Ongoing
   ...
else
```

src/Judge/NTS.Judge.Blazor/Core/Rankings/Protocols/ProtocolRow.razor.cs
achobanov: Or you could assign empty space on the _rankText by default instead null. Down the line we'll want `NText` compoents that might encapsulate such logic, but thats outside of this scope.

src/Judge/NTS.Judge/Core/Behinds/Adapters/RanklistBehind.cs
```diff
@@ -100,9 +102,17 @@ async Task SafeSelectRanking(int id)
...
 
+    async void CreateDocument(Ranking ranking)
+    {
         Ranklist = new Ranklist(ranking);
-        EmitChange();
+        var enduranceEvent = await _events.Read(0);
+        GuardHelper.ThrowIfDefault(enduranceEvent);
+        var officials = await _officials.ReadAll();
+        Document = new RanklistDocument(Ranklist!, enduranceEvent, officials);
     }
```
achobanov: This is better, but that's not "exposing" a method. Exposing means you open something up to consumers i -i.e. make it public. My idea was that you can have this public method in the Protocol component where you will Create a document and print it. Instead of exposing a document property you will expose a method to create one

src/Not/Not.Blazor/Components/NPrintableComponent.cs
```diff
@@ -19,6 +19,8 @@ public static void OnToggle(Action handler)
 
     protected bool IsButtonVisible { get; private set; }
 
+    protected string PrintingClassName { get; set; } = string.Empty;
```
achobanov: What's the purpose of this? Can you not use Class instead?

src/Judge/NTS.Judge.Blazor/Core/Rankings/Ranklists/RanklistTable.razor
```diff
@@ -43,9 +43,9 @@
 </RanklistTableFrame>
 
 
-@if (Behind.Ranklist != null)
+@if (Behind.Document!.Ranklist != null)
```
achobanov: Yes this is most definitelly not ideal as you identified correctly

src/Judge/NTS.Judge.Blazor/Core/Rankings/Protocols/Protocol.razor.cs
```diff
@@ -1,9 +1,25 @@
...
+    public RanklistDocument Document { get; set; } = default!;
+
+    protected override void OnParametersSet()
+    {
+        _documentSubtitle =
+            Document.Header.PopulatedPlace.City
+            + "   "
+            + Document.Header.EventSpan
+            + "   "
+            + Document.Header.PopulatedPlace.Country;
```
achobanov: You should not use hardcoded empty spaces for layouting, because it will not scale accordingly.

achobanov (changes requested):

src/Not/Not.Blazor/Components/NPrintableComponent.cs
StoyanGoranov: using `Class` will hide inherited property from NComponent

src/Judge/NTS.Judge.Blazor/Core/Rankings/Protocols/Protocol.razor
```diff
@@ -5,7 +5,7 @@
 <MudItem Class="d-flex flex-row justify-space-around">
     <MudImage Class="pb-5" Height="150" Width="135" Src="images/logo-fei.png" />
     <MudItem Class="d-flex flex-column justify-center">
-        <NHeadline2 Content="@Document.Header.Title" Class="mb-4" />
+        <NHeadline2 Content="@_document!.Header.Title" Class="mb-4" />
```
achobanov: Handle nullable

src/Judge/NTS.Judge.Blazor/Core/Rankings/Protocols/ProtocolRow.razor
```diff
@@ -23,7 +23,7 @@
 
     <ParticipationTable Number="_combination.Number" Phases="_phases" AlignVertically="true" />
 
-    @if (_total != null && !Entry.Participation.IsEliminated())
+    @if (Entry.Participation.IsComplete() && _total != null)
```
achobanov: Remove _total

src/Judge/NTS.Judge.Blazor/Core/Rankings/Ranklists/RanklistTable.razor
```diff
@@ -2,7 +2,7 @@
 
 @inherits NComponent
 
-<NHeadline2 Content="@Behind.Document!.Ranklist!.Title" Class="printable mb-4" />
+<NHeadline2 Content="@Behind.Ranklist!.Title" Class="printable mb-4" />
```
achobanov: Put in if check

src/Judge/NTS.Judge/Core/Behinds/Adapters/RanklistBehind.cs
```diff
@@ -93,6 +92,14 @@ public async Task ExportFei()
         await File.WriteAllTextAsync(path, xml);
     }
 
+    public async Task<RanklistDocument> CreateDocument()
+    {
+        var enduranceEvent = await _events.Read(0);
+        GuardHelper.ThrowIfDefault(enduranceEvent);
+        var officials = await _officials.ReadAll();
+        return new RanklistDocument(Ranklist!, enduranceEvent, officials);
```
achobanov: Handle null

achobanov (changes requested):

src/Judge/NTS.Judge.Blazor/Core/Rankings/Protocols/Protocol.razor.cs
achobanov: Still viable

src/Not/Not.Blazor/Components/NPrintableComponent.cs
achobanov: Remove if unused

achobanov (approved):

PR #469, opened 2025-06-21
Add Logo as Judge icon

StoyanGoranov:


src/Judge/NTS.Judge.MAUI/App.xaml.cs
```diff
@@ -8,8 +8,17 @@ public partial class App : Microsoft.Maui.Controls.Application
...
+#if WINDOWS
+        Microsoft.Maui.Handlers.WindowHandler.Mapper.AppendToMapping(
+            nameof(IWindow),
+            (handler, view) =>
+            {
+                var nativeWindow = handler.PlatformView;
+                nativeWindow.Title = "NTS Judge";
+            }
+        );
+#endif
```
achobanov: Can't we solve this with a platform-specific service?

achobanov (changes requested):

src/Directory.Build.props
```diff
@@ -10,7 +10,8 @@
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageReference> -->
-    <Analyzer Include="$(SolutionDir)/../analyzers/build/Not.Analyzers.dll" />
+    <Analyzer Include="$(SolutionDir)/../analyzers/build/Not.Analyzers.dll"
+			  Condition="Exists('$(SolutionDir)../analyzers/build/Not.Analyzers.dll')" />
```
achobanov: This is wrong. SolutionDir should be passed as an argument in the build script, instead of having it's checks being conditional

src/Judge/NTS.Judge.MAUI/NTS.Judge.MAUI.csproj
```diff
@@ -12,10 +12,10 @@
     <LangVersion>preview</LangVersion>
     
     <!-- Display name -->
-    <ApplicationTitle>EMS.Judge</ApplicationTitle>
+    <ApplicationTitle>NTS Judge</ApplicationTitle>
 
     <!-- App Identifier -->
-    <ApplicationId>com.es.ems.judge</ApplicationId>
+    <ApplicationId>com.nts.judge</ApplicationId>
```
achobanov: Let's keep this `com.ems.judge`, since it will otherwise messup the app deployment to stores

src/Judge/NTS.Judge.MAUI/installer-script.iss
```diff
@@ -0,0 +1,132 @@
+#define AppExe ".\\bin\\Release\\net8.0-windows10.0.19041.0\\win-x64\\publish\\NTS.Judge.MAUI.exe"
```
achobanov: See if these paths can use wildcards in attempt to remove the specific windows build from the script

achobanov (changes requested):

achobanov:
Furthermore - look into creating the installer using a GitHub Action instead of a local bash script, same as the EMS app.

src/Directory.Build.props
StoyanGoranov: Installer script removed

src/Judge/NTS.Judge.MAUI/installer-script.iss
StoyanGoranov: Installer script removed.

achobanov (dismissed):

PR #481, opened 2025-10-24
Implement Witness RPC client interfaces and dummy data procedures

StoyanGoranov:
## Summary
- add Witness RPC client interfaces mirroring Warp ACL procedures using domain core and watcher types
- implement a WitnessRpcClient backed by DummyData to seed startlist, participant, and snapshot collections
- expose a simple behind class hook to initialize the new RPC client state

## Testing
- not run (dotnet CLI not available in container)

------
https://chatgpt.com/codex/tasks/task_e_68f8fd0ddf388328875caac0508ad8a9


src/Witness/NTS.Witness/RPC/Procedures/IWitnessClientProcedures.cs
```diff
@@ -0,0 +1,3 @@
+namespace NTS.Witness.RPC.Procedures;
+
+public interface IWitnessClientProcedures : IWitnessParticipantsClientProcedures, IWitnessStartlistClientProcedures { }
```
achobanov: Do we need this unbrella interface?

src/Witness/NTS.Witness/RPC/Procedures/IWitnessHubProcedures.cs
```diff
@@ -0,0 +1,6 @@
+namespace NTS.Witness.RPC.Procedures;
+
+public interface IWitnessHubProcedures
```
achobanov: I don't think we need this one as well,

src/Witness/NTS.Witness/RPC/WitnessRpcClient.cs
```diff
@@ -0,0 +1,179 @@
...
+        if (!_startlists.TryGetValue(phaseNumber, out var startlist))
+        {
+            startlist = new WitnessStartlist();
+            _startlists[phaseNumber] = startlist;
+        }
+
+        return startlist;
+    }
+
+    static void UpdateCollection<T, TKey>(
```
achobanov: We should drop domain records, which will enable us to rely on consistent Equals behavior for domain objects. Once we have consistency we can create an extension method on `ICollection` for example which modifies the collection based on `action` type

PR #482, opened 2025-11-22
Port Witness to Blazor server

StoyanGoranov:
You must register User Secrets to review the new Auth implementation. Here are the steps:

1. Run `dotnet user-secrets set "Authentication:Google:ClientId" "YOUR_CLIENT_ID" --project Witness\NTS.Witness.Web` 
1. Run  `dotnet user-secrets set "Authentication:Google:ClientSecret" "YOUR_ACTUAL_SECRET" --project Witness\NTS.Witness.Web`


src/Not/Not/Authentication/AuthModel.cs
```diff
@@ -0,0 +1,23 @@
...
+}
+
+public class AuthConfig
+{
+    public List<AuthUser> Users { get; set; } = [];
+}
+
+public static class Auth
+{
+    public static AuthConfig GetAuthConfigFromAppSettings(WebApplicationBuilder builder)
```
achobanov: Why does this need a builder reference? This locks the implementation to work with Web applications only. Instead it should be an extension of the `IConfiguration` class

src/Not/Not/Authentication/AuthModel.cs
```diff
@@ -0,0 +1,23 @@
+﻿using Microsoft.AspNetCore.Builder;
+using Microsoft.Extensions.Configuration;
+
+namespace Not.Authentication;
+
+public class AuthUser
```
achobanov: Let's use `NUser`

src/Not/Not/Authentication/AuthModel.cs
```diff
@@ -0,0 +1,23 @@
...
+{
+    public string Email { get; set; } = string.Empty;
+    public string[] Roles { get; set; } = [];
+}
+
+public class AuthConfig
```
achobanov: `NAuthenticationSettings`. Also lets split them into files.

src/Not/Not/Authentication/GmailAuth.cs
```diff
@@ -0,0 +1,72 @@
...
+        WebApplicationBuilder builder
+    )
+    {
+        var authConfig = Auth.GetAuthConfigFromAppSettings(builder);
+        var allowedUsersByEmail = authConfig
+            .Users.Where(u => !string.IsNullOrWhiteSpace(u.Email))
+            .ToDictionary(u => u.Email, StringComparer.OrdinalIgnoreCase);
+
+        authBuilder.AddGoogle(options =>
+        {
```
achobanov: Isn't the logic within this method mostly going to be copy-paste when we add move providers?

src/Not/Not/Authentication/GmailAuth.cs
```diff
@@ -0,0 +1,72 @@
+﻿using System.Security.Claims;
+using Microsoft.AspNetCore.Authentication;
+using Microsoft.AspNetCore.Authentication.OAuth;
+using Microsoft.AspNetCore.Builder;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Not.Authentication;
+
+public static class GmailAuth
```
achobanov: This is not a good class name. Called `Gmail`, but has to do with `Google` account. It's not consistent at all with how we name static methods and in general is not very indicative to it's identity.

src/NTS.Blazor/Components/ParticipationChips/ParticipationChipsBehind.cs
```diff
@@ -3,4 +3,15 @@
 
 namespace NTS.Blazor.Components.ParticipationChips;
 
-public class ParticipationChipsBehind : NChipsBase<Participation> { }
+public class ParticipationChipsBehind : NChipsBase<Participation>
+{
+    protected Action<Participation> _selectAction = default!;
```
achobanov: Should not be underscored and shoudl be a Property, not a field.

src/NTS.Blazor/Components/ParticipationChips/ParticipationChips.razor
```diff
@@ -17,7 +17,7 @@
         {
             var number = participation.Combination.Number;
             var color = SetChipColor?.Invoke(participation) ?? Color.Primary;
-            <MudChip Text="@number.ToString()" Value="@participation" Color="@color" Style="min-width: 1rem" Size="ChipSize"  />
+            <MudChip Text="@number.ToString()" Value="@participation" Color="@color" OnClick="@(() => _selectAction(participation))" Style="min-width: 1rem" Size="ChipSize"  />
```
achobanov: Isn't that going to cause null reference if it's not provided as parameter?

src/NTS.Blazor/Components/ParticipationChips/ParticipationChipsBehind.cs
achobanov: Also why does it exist at all. Why not use the `SelectAction` instead? Also if it is only assinged in case `SelectAction` is not null, why is it defined in order to suppress the "Possible null reference error"?

src/Not/Not.Blazor/Components/Base/NChipsBase.cs
```diff
@@ -12,4 +12,7 @@ public class NChipsBase<T> : NBindableComponent<T>
 
     [Parameter]
     public Func<T, Color>? SetChipColor { get; set; }
+
+    [Parameter]
+    public Action<T>? SelectAction { get; set; }
```
achobanov: Rename to `OnSelect`

src/Witness/NTS.Witness.Blazor/Components/Layout/NavMenu.razor
```diff
@@ -0,0 +1,25 @@
...
+@using Not.Blazor.Components
+@using Not.Blazor.Navigation
+
+@inherits NavMenuBehind
+
+<MudNavMenu>
+    <AuthorizeView Roles="admin,official">
+        <Authorized>
+            <NNavLink Icon="@Icons.Material.Filled.Timer" Endpoint="@SNAPSHOT" AfterNavigation="AfterNavigationAction">@Snapshot_page_string</NNavLink>
+            <NNavLink Icon="@Icons.Material.Filled.Checklist" Endpoint="@STARTLIST" AfterNavigation="AfterNavigationAction">@Startlist_string</NNavLink>
```
achobanov: What's this `AfterNavigationAction`?

src/Witness/NTS.Witness.Blazor/Components/Layout/NavMenuBehind.cs
```diff
@@ -0,0 +1,9 @@
+﻿using Not.Blazor.Components;
+
+namespace NTS.Witness.Blazor.Components.Layout;
+
+public class NavMenuBehind : NComponent
+{
+    [Parameter]
+    public Action AfterNavigationAction { get; set; } = default!;
```
achobanov: It should be either `Required` or nullable

src/Witness/NTS.Witness.Blazor/Components/Pages/Performance.razor
```diff
@@ -1,6 +1,8 @@
-﻿@using NTS.Blazor.Components.ParticipationTable
+﻿@using Microsoft.AspNetCore.Authorization
+@using NTS.Blazor.Components.ParticipationTable
 @using NTS.Domain.Core.Aggregates.Participations
 @attribute [Route(PERFORMANCE)]
+@attribute [Authorize(Roles = "admin,competitor")]
```
achobanov, line 5: Extract roles in constants

src/Witness/NTS.Witness.Blazor/WitnessBlazorRootBehind.cs
```diff
@@ -0,0 +1,35 @@
...
+    protected IEnumerable<Assembly> _routeAssemblies = [typeof(WitnessBlazorRoot).Assembly];
+    protected NErrorBoundary _errorBoundary = default!;
+
+    [Inject]
+    public IEnumerable<IStartupInitializer> Initializers { get; set; } = default!;
+
+    [Inject]
+    public IRpcSocket RpcSocket { get; set; } = default!;
+
+    [Parameter]
```
achobanov: Required or null

src/Witness/NTS.Witness.MAUI/NTS.Witness.MAUI.csproj
```diff
@@ -0,0 +1,102 @@
...
+      <ProjectReference Include="..\..\NTS\NTS.csproj" />
+    </ItemGroup>
+
+    <ItemGroup>
+      <MauiXaml Update="SplashPage.xaml">
+        <Generator>MSBuild:Compile</Generator>
+      </MauiXaml>
+    </ItemGroup>
+
+    <!--<ItemGroup>
```
achobanov: Is this needeD?

src/Witness/NTS.Witness.Web/Properties/launchSettings.json
```diff
@@ -0,0 +1,39 @@
+{
+  "profiles": {
```
achobanov, line 2: Use the same launch profiles pattern we use for Judge, Nexus - Localhost, Staging etc

src/Witness/NTS.Witness.Web/Program.cs
```diff
@@ -0,0 +1,81 @@
...
+        builder
+            .Services.AddAuthentication(options =>
+            {
+                options.DefaultScheme = CookieAuthenticationDefaults.AuthenticationScheme;
+                options.DefaultChallengeScheme = GoogleDefaults.AuthenticationScheme;
+            })
+            .AddCookie()
+            .AddGmailAuth(builder);
+
+        builder.Services.AddAuthorization();
```
achobanov: Those should be encapsulated in the `AddWitnessServices`

src/Witness/NTS.Witness.Web/Program.cs
```diff
@@ -0,0 +1,81 @@
...
+        // Endpoint to sign out
+        app.MapGet(
+            "/signout",
+            async (HttpContext ctx) =>
+            {
+                await ctx.SignOutAsync();
+                ctx.Response.Redirect("/profile");
+            }
+        );
+        app.MapGet("/access-denied", () => Results.Text("Access denied."));
```
achobanov: Endpoint mapping should be separated from Program.cs

achobanov (changes requested):

src/Witness/NTS.Witness.Blazor/Components/Layout/NavMenu.razor
StoyanGoranov: NavMenu is located in MudDrawer which remains open after navigation. We need to pass an AfterNavigationAction method to close it.

achobanov:
Steps necessary to start to project should be in a Readme.db. You can create one in the Witness root directory

src/Witness/NTS.Witness.Blazor/Components/Layout/NavMenu.razor
achobanov: Ok, lets rename to `OnNavigation`

src/Witness/NTS.Witness.Web/Program.cs
achobanov: They shounld't also be in the `NTS.Witness` project. Which project is responsible for web stuff?

src/Not/Not/Authentication/User/NUserResolution.cs
```diff
@@ -0,0 +1,46 @@
+﻿using System.Security.Claims;
+using Microsoft.AspNetCore.Authentication;
+
+namespace Not.Authentication.User;
+
+public class NUserResolution : IUserResolver
```
achobanov: Implementations should be named after the implemented interfaces - `NUserResolver`

src/Not/Not/Authentication/User/NUserResolution.cs
```diff
@@ -0,0 +1,46 @@
...
+            ClaimTypes.Role
+        );
+
+        // clear old identities and add new one
+        context.Principal = new ClaimsPrincipal(newIdentity);
+
+        // deny if no email / not gmail / not in allow list
+        if (
+            string.IsNullOrWhiteSpace(email)
+            || !email.EndsWith("@gmail.com", StringComparison.OrdinalIgnoreCase)
```
achobanov: Why are we enforcing only gmail emails?

src/Not/Not/Authentication/User/NUserResolution.cs
```diff
@@ -0,0 +1,46 @@
...
+            context.Fail("Not allowed");
+            context.HandleResponse();
+            return Task.CompletedTask;
+        }
+
+        foreach (var role in authUser.Roles ?? [])
+        {
+            if (!string.IsNullOrWhiteSpace(role))
+            {
+                newIdentity!.AddClaim(new Claim(ClaimTypes.Role, role));
```
achobanov: Why is `!` necessary?

src/Not/Not/Authentication/User/NUserResolution.cs
```diff
@@ -0,0 +1,46 @@
...
+using Microsoft.AspNetCore.Authentication;
+
+namespace Not.Authentication.User;
+
+public class NUserResolution : IUserResolver
+{
+    public Task UserResolution(TicketReceivedContext context, Dictionary<string, NUser> allowedUsersByEmail)
+    {
+        var email = context.Principal?.FindFirst(ClaimTypes.Email)?.Value;
+        var oldIdentity = (ClaimsIdentity)context.Principal!.Identity!;
```
achobanov: Are we sure that `Principal` and `Identity` are never null?

src/Not/Not/Authentication/AuthenticationBuilderExtensions.cs
```diff
@@ -0,0 +1,40 @@
...
+
+public static class AuthenticationBuilderExtensions
+{
+    public static AuthenticationBuilder AddGoogleAuth(
+        this AuthenticationBuilder authBuilder,
+        IConfiguration configuration
+    )
+    {
+        authBuilder.Services.AddSingleton<IUserDeserializer, NAuthenticationSettings>();
+        authBuilder.Services.AddSingleton<IUserResolver, NUserResolution>();
```
achobanov: Are these specific to Google?

src/Not/Not/Authentication/AuthenticationBuilderExtensions.cs
```diff
@@ -0,0 +1,40 @@
...
+
+        authBuilder.AddGoogle(options =>
```
achobanov: Looks better, but still stuff to be extracted - implement a second provider and see the similarities, then improve the abstraction

src/Not/Not/Authentication/NAuthenticationSettings.cs
```diff
@@ -0,0 +1,17 @@
+﻿using Microsoft.Extensions.Configuration;
+using Not.Authentication.User;
+
+namespace Not.Authentication;
+
+public class NAuthenticationSettings : IUserDeserializer
+{
+    public List<NUser> Users { get; set; } = [];
+
+    public Dictionary<string, NUser> GetAllowedUsers(IConfiguration configuration)
```
achobanov: Using dictionary  in the abstraction will work for now, but is not a good approach - what happens if you have 10 mln users? Are you going to load all of them in-memory just to log one? Instead the abstraction should be dynamic - get user by email. The current provider can still load all users from appsettings, since they are few, but that way the abstraction does not have to change when we move to a database provider

src/Not/Not/Authentication/NAuthenticationSettings.cs
```diff
@@ -0,0 +1,17 @@
...
+    {
+        Users = configuration.GetSection("Auth:Users").Get<List<NUser>>() ?? [];
```
achobanov: While this works, it's not very convinient and callers will have to supply the settings with `IConfiguration` instances every time. Instead the settings class should be registered as a singleton object, which is bound to the relevat IConfiguration section during service configuration: [details here](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/configuration/?view=aspnetcore-10.0#bind-hierarchical-configuration-data-using-the-options-pattern)

src/NTS/NtsServiceCollectionExtensions.cs
```diff
@@ -20,4 +23,22 @@ public static IServiceCollection ConfigureNtsBlazor(this IServiceCollection serv
...
+        IConfiguration configuration
+    )
+    {
+        services
+            .AddAuthentication(options =>
+            {
+                options.DefaultScheme = CookieAuthenticationDefaults.AuthenticationScheme;
+                options.DefaultChallengeScheme = GoogleDefaults.AuthenticationScheme;
+            })
+            .AddCookie()
```
achobanov: This should be encapsulated in the `Not.Authentication` and the authentication builder instance should be provided in the same way as the `options => { ... }` argument above. This approach encapsulates stuff that the caller doesn't need to know such as `AddCookie` and `AddAuthorization`, but still allows the caller to configure different authentication providers.

achobanov: Furthermore the entire method should be in the `NTS.Witness` for now. Eventually we'll probably extract a separate service to allow centralized auth for all applications

src/Witness/NTS.Witness.Web/Properties/launchSettings.json
achobanov: `iisSettings` isn't necessary.

achobanov (changes requested):
Getting better, but still needs work. Don't forget unresolved past comments

PR #487, opened 2026-01-04
Update workflows

achobanov:
This PR updates the paths of workflows ahead of merging the restructure


achobanov:
Checks are expected to fail, due to the wrong paths

PR #490, opened 2026-01-05
Implement NTS Witness SignalR communication

StoyanGoranov:


nugets/Not.Application/RPC/IRpcContext.cs
```diff
@@ -0,0 +1,12 @@
+﻿using Not.Domain.Aggregates;
+using Not.Injection;
+
+namespace Not.Application.RPC;
```
achobanov: This has no place in Not. namespace, because it's specific to NTS.

nugets/Not.Application/RPC/RpcSocketExtensions.cs
```diff
@@ -41,18 +41,22 @@ T2 parameter2
         }
     }
 
-    public static async Task<RpcInvokeResult<T>> InvokeOutputProcedure<T>(this IRpcSocket socket, string name)
+    public static async Task<RpcInvokeResult<T1>> InvokeOutputProcedure<T1, T2>(
```
achobanov: This is no longer just output procedure. Also the generic arguments should imply their purpose - `T` and `TResult` for example.

src/Apps/Judge/NTS.Judge.Blazor/Core/Dashboards/Actions/Eliminations/IEliminations.cs
achobanov, line 1: Why was this changed?

src/Apps/Judge/NTS.Judge.MAUI/MauiProgram.cs
```diff
@@ -15,9 +15,9 @@ public static MauiApp CreateMauiApp()
     {
         var builder = MauiApp.CreateBuilder();
         builder.UseMauiApp<App>().ConfigureFonts(fonts => fonts.AddFont("OpenSans-Regular.ttf", "OpenSansRegular"));
-        builder.UseNLog().AddFilesystemLogger();
 
         builder.Services.ConfigureJudgeMaui(builder.Configuration);
+        builder.UseNLog().AddFilesystemLogger();
```
achobanov: ?

src/Apps/Warp/NTS.Warp/Features/Judge/Procedures/IJudgeHubProcedures.cs
```diff
@@ -0,0 +1,3 @@
+﻿namespace NTS.Warp.Features.Judge.Procedures;
+
+public interface IJudgeHubProcedures : IParticipationHubProcedures { }
```
achobanov: Is this necessary? Seems like you can simply rename `IParticipationHubProcedures` -> `IJudgeHubProcedures`

src/Apps/Warp/NTS.Warp/Features/Witness/Procedures/IWitnessClientProcedures.cs
```diff
@@ -0,0 +1,3 @@
+﻿namespace NTS.Warp.Features.Witness.Procedures;
+
+public interface IWitnessClientProcedures : IParticipantsClientProcedures, IStartlistClientProcedures { }
```
achobanov: Is this necessary?

src/Apps/Warp/NTS.Warp/Features/Witness/Procedures/IParticipantsClientProcedures.cs
```diff
@@ -0,0 +1,9 @@
+using Not.Collections;
+using NTS.Domain.Core.Aggregates;
+
+namespace NTS.Warp.Features.Witness.Procedures;
+
+public interface IParticipantsClientProcedures
+{
+    Task ReceiveParticipation(Participation participation, NCollectionAction action);
```
achobanov: `Receive` is good enough name IMO

src/Apps/Warp/NTS.Warp/Features/Witness/Procedures/IStartlistClientProcedures.cs
```diff
@@ -0,0 +1,9 @@
+using Not.Collections;
+using NTS.Domain.Core.Objects.Startlists;
+
+namespace NTS.Warp.Features.Witness.Procedures;
+
+public interface IStartlistClientProcedures
+{
+    Task ReceiveStartlistEntry(StartlistEntry entry, NCollectionAction action);
```
achobanov: `Receive` is good enough name IMO

src/Apps/Warp/NTS.Warp/Features/Witness/Procedures/IWitnessHubProcedures.cs
```diff
@@ -0,0 +1,14 @@
...
+using NTS.Application.Models;
+using NTS.Domain.Aggregates;
+using NTS.Domain.Core.Aggregates;
+using NTS.Domain.Core.Objects.Startlists;
+
+namespace NTS.Warp.Features.Witness.Procedures;
+
+public interface IWitnessHubProcedures
+{
+    Task<IEnumerable<Participation>> SendParticipants(WarpRequest request);
```
achobanov: `SendParticipations`

src/Apps/Warp/NTS.Warp/Features/Witness/WitnessRpcHub.cs
```diff
@@ -47,38 +47,44 @@ public override async Task OnDisconnectedAsync(Exception? exception)
         {
             return;
         }
-        await judgeClient.OnWitnessDisconnedted(Context.ConnectionId);
+        await judgeClient.OnWitnessDisconnected(Context.ConnectionId);
     }
 
-    public async Task<Dictionary<int, EmsStartlist>> SendStartlist(WarpRequest request)
+    public async Task<Startlist?> SendStartlist(WarpRequest request)
```
achobanov: I don't think we should couple Warp with domain entities. it was like that before just to keep the individual applicaitons NTS.Judge and EMS.Witness clean of each other's models. However now we have a single unified model. Warp should just proppagate the models and the Startlist should be constructed on the Witness end.

achobanov: Applies to both Hubs

src/Apps/Witness/NTS.Witness.Blazor/Components/EnduranceEvents/EnduranceEvents.razor
```diff
@@ -0,0 +1,15 @@
+﻿@using NTS.Domain.Setup.Aggregates
+@inherits EnduranceEventsBehind
+
+<NSimpleTable T="UpcomingEvent" Items="@Events" EmptyMessage="@Empty_event_list_message" Height="30vh" Headings="@EventsTableHeaders">
+    <CustomCell>
+        @if (IsConnected && Selected.Event == context)
+        {
+            <MudButton Variant="Variant.Filled" Color="Color.Error" Size="Size.Small" OnClick="() => Disconnect()">@Disconnect_string</MudButton>
```
achobanov: Use our standard Primary & Secondary buttons

src/Apps/Witness/NTS.Witness.Blazor/Components/EnduranceEvents/EnduranceEventsBehind.cs
```diff
@@ -0,0 +1,48 @@
+﻿using Not.Application.RPC;
+using NTS.Application.Warp;
+using NTS.Domain.Setup.Aggregates;
+using NTS.Witness.Services;
+
+public class EnduranceEventsBehind : ComponentBase
```
achobanov, line 8: should be `NComponent`

src/Apps/Witness/NTS.Witness.Blazor/Components/Performances/Performance.razor
```diff
@@ -1,13 +1,40 @@
...
+@attribute [Authorize(Roles = $"{ADMIN},{OFFICIAL},{COMPETITOR}")]
 @inherits PerformanceBehind
 
 <MudStack AlignItems="AlignItems.Center">
-    <NHeadline3 Content="@Participant!.ToString()"/>
-
-    <ParticipationTable Number="1" Phases="PhasesCollection" Editable="false" />
+    <AuthorizeView Roles="@($"{ADMIN},{OFFICIAL}")">
+        <Authorized>
+            <MudGrid Class="d-flex justify-center">
```
achobanov: Do we really need grid with flex for single Select item?

src/Apps/Witness/NTS.Witness.Blazor/Components/Performances/Performance.razor
```diff
@@ -1,13 +1,40 @@
...
+            <ParticipationTable Class="mb-10" Phases="Participant.Phases" Editable="false" />
+    }
+    else
+    {
+        <AuthorizeView Roles="@($"{COMPETITOR}")">
+            <Authorized>
+                <NHeadline3 Class="mx-10 mt-20" Content="@Empty_performance_competitor_message"/>
+            </Authorized>
+        </AuthorizeView>
+        <AuthorizeView Roles="@($"{ADMIN},{OFFICIAL}")">
```
achobanov: What happens if someone has all 3 roles?

src/Apps/Witness/NTS.Witness.Blazor/Components/Performances/PerformanceBehind.cs
```diff
@@ -9,17 +11,34 @@ public class PerformanceBehind : NComponent
     [Inject]
     IPerformanceService Service { get; set; } = default!;
 
-    protected List<Phase> Phases { get; set; } = [];
-    protected PhaseCollection? PhasesCollection { get; set; }
-    protected Person? Participant { get; set; }
+    protected List<NotListModel<Person>> People { get; set; } = [];
+    protected Person SelectedPerson { get; set; } = default!;
+    protected Participation? Participant { get; set; } = default!;
```
achobanov: Participation or Participant?

src/Apps/Witness/NTS.Witness/Services/IEmergencyService.cs
```diff
@@ -0,0 +1,6 @@
+﻿namespace NTS.Witness.Services;
+
+public interface IEmergencyService
```
achobanov: Service is something that contains logic. Context is a better name for classes containing state.

src/Apps/Witness/NTS.Witness/Services/ParticipationService.cs
```diff
@@ -1,35 +1,55 @@
...
     {
-        _participations.Update(participation, action);
+        return Task.FromResult(true);
     }
 
-    public List<Participation> GetParticipations()
+    public void Update(Participation participation, NCollectionAction action)
     {
-        return DummyData.CreateParticipations(10);
+        Participations.ToList().Update(participation, action);
```
achobanov: This line is doing nothing atm.

src/Apps/Witness/NTS.Witness/Services/ParticipationService.cs
```diff
@@ -1,35 +1,55 @@
...
+        if (action == NCollectionAction.AddOrUpdate)
+        {
+            State.AddOrReplace(participation);
+        }
+        else if (action == NCollectionAction.Remove)
+        {
+            State.Remove(participation);
+        }
```
achobanov: Instead of having this here - Add a method in `ObservableList` that handles `NCollectionAction`

src/Apps/Witness/NTS.Witness/Services/WitnessEvents.cs
```diff
@@ -0,0 +1,20 @@
+﻿using Not.Application.CRUD.Ports;
+using Not.Storage.REST;
+using NTS.Domain.Setup.Aggregates;
+
+namespace NTS.Witness.Services;
+
+public class WitnessEvents : IWitnessEvents
```
achobanov: How is this service useful?

src/Apps/Witness/NTS.Witness/Storage/WitnessStorageBuilder.cs
```diff
@@ -0,0 +1,50 @@
...
+        readonly NStorageBuilder _nStorageBuilder;
+        readonly IServiceCollection _services;
+
+        internal Builder(IServiceCollection services, IConfiguration configuration)
+        {
+            _nStorageBuilder = new(services, configuration);
+            _services = services;
+        }
+
+        public Builder AddMongoStorage(string? connectionString)
```
achobanov: What are you using Mongo for in Witness at this point?

src/Apps/Witness/NTS.Witness/Warp/EventRpcContext.cs
```diff
@@ -0,0 +1,64 @@
...
+
+    public async Task Set(UpcomingEvent upcomingEvent)
+    {
+        if (Event == upcomingEvent)
+        {
+            return;
+        }
+        if (Event != null)
+        {
+            throw new DomainException(Cannot_select_another_event_without_resetting__string, Event);
```
achobanov: This is not a domain model. You can simply call `NotifyHelper.Warn` and return

src/Domain/NTS.Domain.Watcher/SnapshotPayload.cs
```diff
@@ -0,0 +1,23 @@
+﻿using NTS.Domain.Enums;
+
+namespace NTS.Domain.Watcher;
+
+public class SnapshotPayload
```
achobanov, line 5: This is not a good model. 
1. First of all a domain model named `Payload` is wrong, unless it's something to do with a shipping for example. 
2. Additionally it accepts an IntermediateSnapshot, which I know we discussed, but It's not a good solution. The Domain model shouldn't care about transient models which is what this `IntermediateSnapshot looks like. 
3. Finally domain models shouldn't care about parsing string like what's happening here. Instead it should receive values and validate them

Instead you should do as we have done in the other domains - Create a `Snapshot` Aggregate (if it's has to be persisted in the domain context) or a record (value object, `HandoutDocument` for example) and then have a `SnapshotModel` to pass it along

src/NTS.Blazor/Components/ParticipationTable/ParticipationTable.razor.cs
```diff
@@ -34,6 +35,10 @@ protected override void OnParametersSet()
         _anyRepresentation = Phases != null && Phases.Any(x => x.IsReinspectionRequested);
         _anyRequiredInspection = Phases != null && Phases.Any(x => x.IsRequiredInspectionRequested);
         _anyCompulsoryRequiredInspection = Phases != null && Phases.Any(x => x.IsRequiredInspectionCompulsory);
+        if (Number == null)
+        {
+            _numberStyles += "opacity:0;";
```
achobanov: Why would Number be null?

src/Apps/Witness/NTS.Witness/Services/ParticipationService.cs
```diff
@@ -1,35 +1,55 @@
...
+        get => Participations;
+        set
+        {
+            State.Clear();
+            State.AddRange(value);
+        }
+    }
 
-    public void Update(Participation participation, NCollectionAction action)
+    protected override Task<bool> CreateState(params IEnumerable<object> arguments)
```
achobanov: If you're not taking advantage of this state initialization process why do you inherit from `NStatefulService`?

src/Apps/Witness/NTS.Witness/Services/IParticipationService.cs
```diff
@@ -0,0 +1,10 @@
+﻿using Not.Application.Behinds.Adapters;
+using Not.Injection;
+using NTS.Domain.Core.Aggregates;
+
+namespace NTS.Witness.Services;
+
+public interface IParticipationService : IStatefulService
```
achobanov: Context

src/Apps/Witness/NTS.Witness/Services/IParticipationService.cs
```diff
@@ -0,0 +1,10 @@
...
+{
+    IEnumerable<Participation> ActiveParticipations { get; }
```
achobanov: `Active` is enough

achobanov (changes requested):

src/Apps/Judge/NTS.Judge.Blazor/Core/Dashboards/Actions/Eliminations/IEliminations.cs
StoyanGoranov: The interface was duplicated, causing problems in Judge Eliminate/Restore functionality.

src/Apps/Judge/NTS.Judge.MAUI/MauiProgram.cs
StoyanGoranov: `AddFilesystemLogger` uses static class `FileContextHelper` to configure where logs will be stored. For that the `FileContextHelper` needs an internal property `_applicationName` to be set. That property is set by `ConfigureNtsStorage` within `ConfigureJudgeMaui` services registration. Therefore `ConfigureJudgeMaui` should be called before `AddFilesystemLogger`

Otherwise a `property is undefined` runtime error occurs.

src/Apps/Warp/NTS.Warp/Features/Witness/Procedures/IParticipantsClientProcedures.cs
StoyanGoranov: Receive is definitely a better name but SignalR cannot distinguish between method overloads with the same number of parameters. Consequently, having the same name leads to those methods never being invoked by the Hub.

https://stackoverflow.com/questions/15686859/signalr-hub-overloads

src/Apps/Warp/NTS.Warp/Features/Witness/Procedures/IStartlistClientProcedures.cs
StoyanGoranov: https://stackoverflow.com/questions/15686859/signalr-hub-overloads

src/Apps/Witness/NTS.Witness.Blazor/Components/Performances/Performance.razor
StoyanGoranov: The way Authorization is designed currently, it doesn't support having multiple roles. Instead roles are ordered in the following hierarchy Admin > Official, Vet > Competitor. The idea is to not show UI that is outside of the role's concern. This also enables adjusting UX to the specifics of the role. For example Emergency Contacts page can be accessed from NavMenu for all roles except Competitor. For Competitors that navigation appears as a floating red dial button on the screen at all times, so that if an emergency arises, Contacts would be very easily accessible. Assigning multiple roles to users doesn't make sense in that approach. If functions availability needs to change, it is best to allow Admin to reassign roles with lesser authority.

src/NTS.Blazor/Components/ParticipationTable/ParticipationTable.razor.cs
StoyanGoranov: Made Number nullable, because in Performance component the Combination's number is already shown by `<NHeadline3 Content="@Participation.ToString()"/>` adding it again would add unnecessary UI clutter.

Alternatively, if Number really is required we can add a HideNumber parameter, however it doesn't seem like Number having a value is essential for the component to function.

src/Apps/Warp/NTS.Warp/Features/Witness/Procedures/IParticipantsClientProcedures.cs
achobanov: I see.

src/Apps/Judge/NTS.Judge.MAUI/MauiProgram.cs
achobanov: Return it, that issue should be resolved in my PR and this will cause conflicts

src/Apps/Witness/NTS.Witness.Blazor/Components/Performances/Performance.razor
achobanov: Usually roles are additive - i.e. permissions are granted to a user by assigning that role to them. This way it's more restrictive and tying up UX feels like a mistake to me, as it will have uninteded consequences. We can keep this for now

src/NTS.Blazor/Components/ParticipationTable/ParticipationTable.razor.cs
achobanov: Feels like unnecessary complexity. Number should either be there or not.

nugets/Not.Application/RPC/RpcSocketExtensions.cs
```diff
@@ -41,22 +41,22 @@ T2 parameter2
         }
     }
 
-    public static async Task<RpcInvokeResult<T1>> InvokeOutputProcedure<T1, T2>(
```
achobanov: <T, TResult> instead of <TResult, T>

src/Apps/Warp/NTS.Warp/Features/Judge/JudgeRpcHub.cs
```diff
@@ -55,9 +60,9 @@ await _witnessRelay
             .Clients.Group(request.EnduranceEventId)
             .ReceiveParticipation(participation, NCollectionAction.AddOrUpdate);
 
-        if (request.Payload.Participation.Phases.Any(x => x.IsComplete()))
+        if (participation.Phases.Any(x => x.IsComplete()))
         {
-            var startlistEntry = new StartlistEntry(request.Payload.Participation);
+            var startlistEntry = new StartlistEntry(participation);
```
achobanov: Change this to Participation as well. No reason for the SignalR hub to know about domain models.

achobanov: We should simply remove domain references from this. However it might be best done separatly after current PRs are mreged to avoid conflicts.

src/Domain/NTS.Domain.Core/Objects/Startlists/Startlist.cs
```diff
@@ -53,6 +53,25 @@ public Startlist(IEnumerable<Participation> participations)
         _history = OrderByTimeThenPhase(history);
     }
 
+    public Startlist(IEnumerable<StartlistEntry> entries)
```
achobanov: Shouldn't be necessary. I'd like the Startlist to work with participations only.

src/Domain/NTS.Domain.Core/Objects/Startlists/StartlistEntry.cs
```diff
@@ -18,11 +19,14 @@ public StartlistEntry(Participation participation)
     {
         Athlete = participation.Combination.Athlete.Names;
         Number = participation.Combination.Number;
-        var currentIndex = participation.Phases.IndexOf(participation.Phases.Current);
-        var nextPhase = participation.Phases[currentIndex + 1];
+        var nextPhase = participation.Phases.GetNext();
         PhaseNumber = participation.Phases.NumberOf(nextPhase);
         Distance = nextPhase.Length;
-        Start = new Timestamp((nextPhase.StartTime ?? Timestamp.DEFAULT).ToDateTimeOffset());
+        if (nextPhase.StartTime == null)
```
achobanov: Revert this to avoid conflicts.

src/NTS.Application/Startlists/StartlistService.cs
```diff
@@ -25,8 +33,29 @@ public StartlistService(IReadMany<Participation> participations, StartlistContex
     protected override async Task<bool> CreateState(params IEnumerable<object> arguments)
     {
         var participations = await _participations.ReadMany();
-        _context.Startlist = new Startlist(participations);
-
+        if (participations.Any())
```
achobanov: If you use Participations instead of StartlistEntries then you'll be able to move this logic in the Witness implementation of `IRepository<Participation>` and keep this service clean

achobanov (changes requested):

src/NTS.Blazor/Components/ParticipationTable/ParticipationTable.razor.cs
StoyanGoranov: Removed the Number param since everywhere the ParticipationTable is used the Number of the Participation is shown outside of the Table as well.

src/NTS.Application/Startlists/StartlistService.cs
StoyanGoranov: This clears up the use of Participation for Startlist initialization everywhere but ends up not initializing Startlist with Setup data in Judge.

src/Apps/Witness/NTS.Witness/Services/ParticipationReader.cs
```diff
@@ -1,22 +1,40 @@
 using System.Linq.Expressions;
 using Not.Application.CRUD.Ports;
 using Not.Injection;
+using NTS.Application.Warp;
 using NTS.Domain.Core.Aggregates;
+using static NTS.Application.Factories.ParticipationAndRankingFactory;
 
 namespace NTS.Witness.Services;
 
 public class ParticipationReader : IReadMany<Participation>, ITransient
```
achobanov, line 10: Filename is incorrect.

achobanov (dismissed):

PR #493, opened 2026-01-18
Refactor Domain models

achobanov:
- AggregateRoot is now `Aggregate` 
- Aggregate is now `Entity` - non-standalone domain entity, encompased in an aggregate
- Equality is done now on `Entity` level using a list of values to check for uniqueness
- Aggregate still defines Id and provides that Id for uniquness checks to `Entity`
- Decoupled `Club`, `Horse` and `Athlete` as _id for example is completely unnecessary for those records in Core domain
- Fixed Nexus Functions project


achobanov (commented):
Domain models ids

PR #497, opened 2026-02-16
Fix signalR Remote Procedure Calls

StoyanGoranov:


src/Domain/NTS.Domain.Core/Objects/Startlists/Startlist.cs
```diff
@@ -92,18 +91,21 @@ public void UpdateState()
     public void Add(Participation participation)
     {
         var index = participation.Phases.IndexOf(participation.Phases.Current);
-        if (participation.Phases.Count <= ++index)
+        if (participation.Phases.Count <= index + 1)
```
achobanov: Good catch, but the more elegant solution and probably what I intended to originally happen is to keep `++index` and remove `+ 1` when accessing next phase below

achobanov (dismissed):

